PR #10, opened 2015-07-08
Add throam.com

frankpape:
Used by http://www.throwawaymail.com/


martenson:
thanks for the contribution!

PR #17, opened 2015-09-25
More hushmail domains

joshtronic:
Also looks like someone manually added some domains, ran it through the
directions on the README and it re-arranged the order a bit


joshtronic:
Also added trashcanmail and some new trashmail domains

martenson:
Thanks for the contribution!

PR #19, opened 2015-10-08
Added dodsi.com - temporary mail domain from http://tempail.com

maciekmm:
Used by http://www.tempail.com/temp-email


martenson:
thanks!

PR #20, opened 2015-12-21
Added domains from temp-mail.org

ketanbhatt:


martenson:
all seem legit disposables; thank you @ketanbhatt !

PR #21, opened 2016-02-19
added some addresses we collected

chebelom:


disposable_email_blacklist.conf
```diff
@@ -157,6 +160,7 @@ centermail.com
 centermail.net
 chammy.info
 cheatmail.de
+chickenkiller.com
```
martenson: this does not look like a disposable email domain - http://chickenkiller.com/

disposable_email_blacklist.conf
```diff
@@ -728,6 +741,7 @@ nomorespamemails.com
 nonspam.eu
 nonspammer.de
 noref.in
+nospam.wins.com.br
```
martenson: com.br is not a valid domain name

disposable_email_blacklist.conf
chebelom: fixed, sorry (:

martenson:
thank you very much @chebelom !

PR #22, opened 2016-02-25
New batch

joshtronic:


martenson:
thank you very much @joshtronic !

PR #29, opened 2016-06-06
Add mail2tor.com service : Anonymous/Tor email service provider

chtitux:


martenson:
Anonymous does not mean disposable. This one also looks non-trivial to set up. If people want to protect their privacy with this effort, they probably have good reasons.

Thank you for your contribution @chtitux , but I will not merge it at this point. My list targets a different beast.

PR #30, opened 2016-06-08
Added new throwawaymail.com domain

joshtronic:
Also dropped a few domains that were filtered out by the whitelist file.


martenson:
thanks @joshtronic !

PR #39, opened 2016-06-17
Simplified PHP code

deguif:


martenson:
great, thanks @deguif

deguif:
We could also use this code instead of the __foreach__:
```php
return in_array($mail_domain, $mail_domains_ko);
```

martenson:
@deguif I haven't worked with PHP for a long time; if you think it is better please PR it, I will be happy to merge

PR #40, opened 2016-06-17
Simplified PHP code again

deguif:
Let a comment to avoid undefined offset.
That was also the case with the previous example.


martenson:
thanks @deguif

PR #45, opened 2016-06-20
Added more domains from mytemp.email

deguif:
See https://mytemp.email/

### DIFF
--- a/disposable_email_blacklist.conf
+++ b/disposable_email_blacklist.conf
@@ -192,6 +192,7 @@ aver.com
 axiz.org
 azcomputerworks.com
 azmeil.tk
+b.kyal.pl
 b1of96u.com
 b2cmail.de
 badgerland.eu
@@ -426,6 +427,7 @@ doiea.com
 dolphinnet.net
 domforfb1.tk
 domforfb18.tk
+domforfb19.tk
 domforfb2.tk
 domforfb23.tk
 domforfb27.tk
@@ -482,9 +484,11 @@ e.coza.ro
 e.ezfill.com
 e.hecat.es
 e.hpc.tw
+e.incq.com
 e.lee.mx
 e.ohi.tw
 e.runi.ca
+e.sino.tw
 e.spr.io
 e.ubm.md
 e3z.de
@@ -588,6 +592,7 @@ extremail.ru
 eyepaste.com
 ezstest.com
 f.fuirio.com
+f.fxnxs.com
 f.hmh.ro
 f4k.es
 facebook-email.cf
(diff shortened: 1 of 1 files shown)


disposable_email_blacklist.conf
```diff
@@ -1544,6 +1560,7 @@ secretemail.de
 secure-mail.biz
 secure-mail.cc
 secured-link.net
+securehost.com.es
```
martenson: This one does not seem as a disposable address host (http://securehost.com.es/)
Where are you getting these?

deguif: It was proposed by https://mytemp.email/ when you ask to create a new disposable email address

deguif: You're right, the host doesn't seem to be a disposable domain host. Maybe the domain was sold to a new owner and they use it now for hosting their website. I'm going to remove this one.

martenson: I see. Could you please drop this one out?

martenson:
Also I am not sure about this, but generally we probably want to avoid putting domains of the third level here. If spammer/disposable service uses somedomain.com they can also easily create <anything_here>.somedomain.com so we should filter on the top domain name.

disposable_email_blacklist.conf
frin: Interesting, the securehost.com.es website doesn't work now (Connection refused).

deguif: But http://www.wifiber.es/ which was the same site.

deguif:
Sounds legit.
In this case we should update the script to check email validity to check only for second domain level and in some cases for third domain level (for domains like *.co.uk or *.com.es, etc...)

deguif:
Here the pseudo PHP code that should handle the new behavior.

```php
<?php
//check-disposable-email.php

if (!isset($argv[1])) {
   echo "You should specify a domain to check\n";
   exit(1);
}

$domain = $argv[1];
$disposableDomains = [
    'disposable.co.uk',
    'disposable.com',
];

if (preg_match('/(?:[^.]+\.)?([^.]*\.[^.]+)$/', $domain, $matches)) {
   $twoLevelDomain = $matches[1];
   $threeLevelDomain = $matches[0] !== $matches[1] ? $matches[0] : null;

   echo 'Two level domain: '.$twoLevelDomain."\n";
   if (null !== $threeLevelDomain) {
       echo 'Three level domain: '.$threeLevelDomain."\n";
   }

   foreach ($disposableDomains as $disposableDomain) {
       if ($disposableDomain === $twoLevelDomain || ($threeLevelDomain && $disposableDomain === $threeLevelDomain)) {
           echo sprintf("Domain %s is a disposable domain.\n", $domain);
           return true;
       }
   }

   echo sprintf("Domain %s is not a disposable domain.\n", $domain);
   return false;
}
```

martenson:
@deguif I would like to leave the implementation details on the deployer to keep this list simple and not opinionated. We can provide hints and pointers though.

PR #52, opened 2016-08-18
Add 4 disposable email domains

wootcake:


wootcake:
stuckmail.com seems to be down right now, however it worked yesterday. Not sure if it's a temporary downtime.

martenson:
Thank you very much @wootcake ! I trust you on the stuckmail.

PR #56, opened 2016-08-23
Add mvrht.com, an alias for 10minutemail.com

marcelstoer:


martenson:
thank you @marcelstoer !

PR #57, opened 2016-08-26
New alias tqoai.com from 10minutemail.net

PanPanayotov:
Added a new domain from 10minutemail.net to the list


martenson:
thanks @PanPanayotov !

PR #59, opened 2016-08-31
new alias stexsy.com from temp-mail.org

PanPanayotov:
I have added new stexsy.com from https://temp-mail.org


martenson:
thank you!

PR #61, opened 2016-09-28
new domain xzsok.com from 10minutemail.net

PanPanayotov:
new domain xzsok.com from 10minutemail.net


martenson:
thanks a bunch @PanPanayotov

PR #62, opened 2016-10-13
Some disposable domains MXing to dnses.ro and mytemp.email

kyledrake:


martenson:
thanks a bunch @kyledrake

PR #66, opened 2016-11-25
add abc....com

CRS-ECHO51:
full address >> abcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijk.com


martenson:
This seems like a bizarre but legit email service provider with registration captcha and paid premium accounts. Thanks for the contribution @CRS-ECHO51 but I believe this does not fit the disposable list.

PR #67, opened 2016-11-29
update blacklist

MitsunChieh:
add two blacklist
iaoss.com
euaqa.com


martenson:
Splendid! Thanks @MitsunChieh !

PR #68, opened 2016-12-06
reject dsiay.com

MitsunChieh:
dsiay.com is a 10 mins email


martenson:
awesome sauce! thanks @MitsunChieh

PR #72, opened 2017-01-31
Added 18 new domains from  temp-mail.org and data from my website

JamesAusten:
Found this list very useful thus far, so here are my contributions


martenson:
Thanks a bunch @JamesAusten !

PR #73, opened 2017-02-03
Remove Naver from disposable email list

marionme:
Naver email domain seems to be frequently used in South Korea, and not as a disposable email provider.


martenson:
That seems correct. Thank you for pointing it out @marionme !

http://www.navercorp.com/en/service/naver.nhn

PR #74, opened 2017-02-13
Remove sibmail.com from disposable email list.

yohanson:
It is a real email service, not a disposable one.
Belongs to Tomtel Internet-provider in Tomsk, Russia.
http://www.tomtel.ru/adds.html


martenson:
Thank you @yohanson !

PR #85, opened 2017-03-29
add hidebox.org

defeated:
we've recently started seeing traffic from hidebox.org and thought it could be included here! /cc @martenson


martenson:
thanks a bunch @defeated

PR #89, opened 2017-04-28
more domains

ereli:
More domains from http://www.yopmail.com/en/


szepeviktor:
Squash on merge!

martenson:
we decided to only cover `yy.zz` format and not `xx.yy.zz`
see https://github.com/martenson/disposable-email-domains/issues/46 for details

ereli:
@martenson I understand what you mean - you don't have want to have duplicates of the same domains, so you want to exclude subdomains.
But what if someone is running a service that provides "free domains names" under their domain?
the typical approach taken by the experts at https://publicsuffix.org/list/ is to include it as a "public suffix".
For example, they've included in their public suffix (which is similar but not identical to TLD) list domains like `dyndns.info`, that provides anyone with an IP address a hostname and DNS services.

They've yet to learn about  fr.nf that is providing this "free domain" service. I might send them a pull request to add `fr.nf` and `xl.cx` if I have time.

For the purpose of disposable emails - adding xl.cx wouldn't be right as not all domains under xl.cx related.

martenson:
@ereli I think the argument here is why any legit email server would be under `xx.yy.zz`.

szepeviktor:
@martenson Imagine this.
You own a domain: martenson.com you use several subdomains like www.martenson.com and mail.martenson.com
On a rainy day you donate one of your subdomains to YOPmail.

Then what should **we** list? martenson.com or donated.martenson.com ?

martenson:
@szepeviktor I would +1 listing `martenson.com` - I understand your point though.

Another argument: Second level domain costs money to maintain. Third level domains you can create thousands for cheap.

ereli:
@martenson but there's a false positive risk with adding `martenson.com` and the likes. in the case of fr.nf and other free hostname providers - the free subdomain service wasn't designed just to provide free subdomains for disposable email providers, it has thousands of users the might have a legitimate need for a hostname and don't want to pay 15$ per year for a second level domain.
image

kyledrake:
How about a new list - high_abuse_potential.conf or dynamic_dns.conf or something like that?

martenson:
In https://github.com/martenson/disposable-email-domains/pull/140 I have implemented a check for public suffixes using the list you linked. The fr.nf domain is not on the list which to me means their owners did not make a minimal effort to prevent this kind of abuse of their service.

I intend to add nf.fr and other domains used by yopmail to the blacklist.

Thanks for the discussion and contribution @ereli.

PR #96, opened 2017-06-22
Removed pp.ua domain as it's an ukrainian regristar for private domain

deguif:
See http://drs.ua/ and http://pp.ua/
See also the wikipedia article https://en.wikipedia.org/wiki/.ua

I also reintroduced `add3000.pp.ua`, `ass.pp.ua`, `get.pp.ua`, `ip4.pp.ua`, `ip6.pp.ua`, `loh.pp.ua`, `mox.pp.ua`, `stop-my-spam.pp.ua`, `yopmail.pp.ua` domains

There're already some subdomains entries for `pp.ua` in the `blacklist`file:
- `web-mail.pp.ua`
- `eml.pp.ua`
- `fake-email.pp.ua`
- `jetable.pp.ua`

See https://email-fake.com/mox.pp.ua/tarjeixsi for the list of all `pp.ua` fake domains.

### DIFF
--- a/disposable_email_blacklist.conf
+++ b/disposable_email_blacklist.conf
@@ -112,6 +112,7 @@ abyssmail.com
 ac20mail.in
 academiccommunity.com
 acentri.com
+add3000.pp.ua
 adiq.eu
 adobeccepdm.com
 adpugh.org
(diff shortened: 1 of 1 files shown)


dailyraisin:
👍

deguif:
There is the same issue with `net.ua`, `zp.ua` domains.
They are valid second level with their own registration system.
We should not black list them.

martenson:
@deguif thank you

PR #97, opened 2017-06-22
Removed pcusers.otherinbox.com as otherinbox.com is already black listed

deguif:
See this line: https://github.com/martenson/disposable-email-domains/blob/master/disposable_email_blacklist.conf#L1424


martenson:
Good catch, thanks @deguif !

PR #98, opened 2017-06-22
Remove subdomains for uu.gl which is already black listed

deguif:
See https://github.com/martenson/disposable-email-domains/blob/master/disposable_email_blacklist.conf#L1976


dailyraisin:
👍

martenson:
Thanks for the cleanup @deguif !

PR #100, opened 2017-07-03
Adding domains from EmailOnDeck.com && slippery.email ++ Nodejs Example

martin-fogelman:


martenson:
I am fine with first two commits, but the other two I am not so sure. JS developers can parse the blacklist into objects so I see no reason to do this for them, and the last commit has list of some known email providers, which can never be even close to up to date, which is the reason why we are constructing blacklist instead.

@martin-fogelman If you could split this PR into the domain additions and then the other parts (with explanation of why we need this and maybe some automatic generation so it stays sync'd) that would be great. 

Thank you for your contributions.

martin-fogelman:
@martenson sorry I never intended the js and public emails to be part of this pull request.  I've removed them and added a node js example to read and/or write the .conf to json or whatever else is needed

martin-fogelman:
also not sure how to get the indents from being so damn large...

martenson:
Thanks @martin-fogelman

PR #101, opened 2017-07-04
add spamgourmet.com

sbusso:


martenson:
spamgourmet.com is [whitelisted](https://github.com/martenson/disposable-email-domains/blob/master/whitelist.conf#L146) since it is not a disposable email provider

sbusso:
@martenson spamgourmet.com by its own definition is a disposable email provider: 'spamgourmet - adresses email gratuites et jetables' translated 'Spamgourmet - free and disposable email addresses'

martenson:
@sbusso I think they are using the word disposable in a different way than this repository. If you have to register/create an account we do not mark it as disposable.

sbusso:
> Après avoir entré et confirmé votre adresse protégée, vous pouvez donner à qui vous voulez des adresses jetables qui s'auto-détruisent. Les adresses jetables sont de la forme

That means this address is temporary and will be automatically destroyed after use.

sbusso:
Anyway it is a disposable email address provider, if you dont want to blacklist that is fine, but I will.

martenson:
@sbusso To me 'disposable' is more related to the process of automatically obtaining a random email address you do not care about after you load a page. The fact you can abandon at any point a free email address that you registered for cannot be changed and is not strong enough sign of disposability for me. If they periodically remove inactive accounts, that is a hint of disposability but not a decisive one (as I see it). Thanks for your input! @sbusso

sbusso:
@martenson That is where I dont get it, because it is exactly what you describe as disposable. Yes you can get a free account but, this free account is to give you access to as much as disposable emails you need.

PR #103, opened 2017-07-11
Adding Node.js Example Email Check

martin-fogelman:
Just realized the other language examples also said how to use it, so I've added a js example using .includes and .split


martenson:
fixed by https://github.com/martenson/disposable-email-domains/pull/100 I believe

PR #120, opened 2017-09-05
Added anyalias.com to the list

harush:


martenson:
I am a bit ambivalent regarding flagging of this service. It requires a registration (good) but then gives you unlimited addresses(bad).

>Sign up and choose your new anyalias username, for example "john".
Need to sign up for a shady website you don't trust with your private email account?
You can now pick any alias as a disposable email address like shady@john.anyalias.com and keep your email safe.

Opinions @deguif ?

deguif:
I would add that we have many services like this one on the whitelist like `spamgourmet`.

martenson:
@deguif true, thanks 👍

martenson:
@harush thanks for the contribution!

PR #126, opened 2017-10-21
C# usage sample

andrewvk:


README.md
```diff
@@ -86,6 +86,30 @@ rl.on('close', () => {
...
+          if (!string.IsNullOrWhiteSpace(line) && !line.TrimStart().StartsWith("//"))
+            result.Add(line);
+      }
+      return result;
+    });
+
+...
+
+var addr = new MailAddress(email);
+if (_emailBlackList.Value.Contains(addr.Host)))
```
deguif: You should compare with top level domain (2 or 3 levels), not the entire domain.

For example, if the disposable email is `test.spamtrap.com`, there will be only `spamtrap.com` in the black list, as we black list the top domain, for avoiding every subdomain to be declared in the file (that could be a huge task an hard to maintain, as it's very easy to create a subdomain).

There are some case where the domain can have three levels (for example, `spamtrap.co.uk`)...


The others langage usage samples are not all updated as the switch to top level domain blacklist is recent.

deguif: See discussion on #45 about the domain levels.

deguif:
Thanks again @andrewvk

PR #132, opened 2017-11-12
More efficient PHP code based on original

pjebs:
- $path defined so the code works from anywhere in a framework folder
structure
- file is loaded to an “associative array” instead of a standard array.
The prior approach had to search through entire array to find email
address. This approach can use the key index.

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,14 +31,16 @@ Available as [PyPI module](https://pypi.python.org/pypi/disposable-email-domains
 True
 ```
 
-**PHP** contributed by @txt3rob and @deguif
+**PHP** contributed by [@txt3rob](https://github.com/txt3rob), [@deguif](https://github.com/deguif) and [@pjebs](https://github.com/pjebs)
 ```php
-function is_temp_mail($mail) {
-    $mail_domains_ko = file('disposable_email_blacklist.conf', FILE_IGNORE_NEW_LINES | FILE_SKIP_EMPTY_LINES);
-
-    //Need to ensure the mail contains an @ to avoid undefined offset
-    return in_array(explode('@', $mail)[1], $mail_domains_ko);
+function is_disposable_email($email) {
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -31,14 +31,16 @@ Available as [PyPI module](https://pypi.python.org/pypi/disposable-email-domains
...
 ```php
-function is_temp_mail($mail) {
-    $mail_domains_ko = file('disposable_email_blacklist.conf', FILE_IGNORE_NEW_LINES | FILE_SKIP_EMPTY_LINES);
-
-    //Need to ensure the mail contains an @ to avoid undefined offset
-    return in_array(explode('@', $mail)[1], $mail_domains_ko);
+function is_disposable_email($email) {
+  $path = realpath(dirname(__FILE__)) . '/disposable_email_blacklist.conf';
+  $mail_domains_ko = file($path, FILE_IGNORE_NEW_LINES | FILE_SKIP_EMPTY_LINES);
+  $mail_domains_ko = array_combine($mail_domains_ko, array_fill(0,count($mail_domains_ko), null));
```
deguif: `array_fill_keys` is better, it will avoid to use `array_fill`. Maybe fill values with `true`, instead of `null`, this will avoid need to use `array_key_exists`, `isset` will be sufficient and is more performant than `array_key_exists`.

Before: `array_combine($mail_domains_ko, array_fill(0,count($mail_domains_ko), null))`
After: `array_fill_keys($mail_domains_ko, true)`

README.md
```diff
@@ -31,14 +31,16 @@ Available as [PyPI module](https://pypi.python.org/pypi/disposable-email-domains
...
+  $domain = mb_strtolower(explode('@', trim($email))[1])
+  return (isset($mail_domains_ko[$domain]) || array_key_exists($domain, $mail_domains_ko));
```
deguif: If array values are not `null`, `isset` is sufficient.

pjebs:
See https://github.com/martenson/disposable-email-domains/pull/133

PR #146, opened 2018-02-28
Link to contributors

di:
Also updates links to https://pypi.org instead, and adds syntax highlighting for the Python example.


martenson:
thanks @di!

PR #150, opened 2018-03-06
Update disposable_email_blacklist.conf

AndrejfSantos:
3 new disposable emails used to register at mobrand.com

### DIFF
--- a/disposable_email_blacklist.conf
+++ b/disposable_email_blacklist.conf
@@ -66,6 +66,7 @@
 20email.it
 20mail.in
 20mail.it
+20mail.eu
 20minutemail.com
 20mm.eu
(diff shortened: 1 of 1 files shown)


deguif:
Thanks @AndrejfSantos 
This PR looks great.

disposable_email_blacklist.conf
```diff
@@ -66,6 +66,7 @@
 20email.it
 20mail.in
 20mail.it
+20mail.eu
```
deguif: Shouldn't it be before `20mail.in`, alphabetically speaking?

AndrejfSantos: indeed it should, will change that

PR #153, opened 2018-04-03
Include a bunch of new email domains

hackerkid:
These are domain from  https://github.com/ivolo/disposable-email-domains/blob/master/index.json.


martenson:
thank you @hackerkid

PR #160, opened 2018-05-14
Add bk.ru, inbox.ru, list.ru, mail.ru to blacklist

kyledrake:
I just cleaned up a massive spam attack with several thousand accounts from these 4 domains. They are either disposable or are a hacking front, I'm not sure which.


deguif:
We can't add mail.ru to the disposable list as it's a major email provider in Russia.

Pascal76:
All theses domaines are from mail.ru (MX is mxs.mail.ru)
ALL ARE LEGITIMS.
 
>They are either disposable or are a hacking front, I'm not sure which

Your thinking/request is silly and dangerous for people who need reliable disposable email domain list.

PR #161, opened 2018-05-17
add nwytg.com

tkrajcar:


deguif:
Thanks @tkrajcar 
I confirm this a disposable as the domain is used by 10minutemail.com

PR #162, opened 2018-05-18
Update disposable_email_blacklist.conf

juanjointera5:
Add disposable domains from https://temp-mail.org/


deguif:
Thanks @juanjointera5

PR #166, opened 2018-06-07
Add freemymail.org and sawoe.com

lodow:


martenson:
@lodow thanks for the PR, what is the source of these domains?

PR #167, opened 2018-06-26
Add maskmail.net

kdeldycke:
https://maskmail.net is an anonymous email addresses redirector.


deguif:
As maskmail requires to register on their website, I'm not sure this should go to blacklist.
ping @martenson

martenson:
I vote for not including this in the blacklist. To my understanding this will not time out the masking email addresses so you have to explicitly dispose of it. This, to me, looks equal to filtering out the emails.

kdeldycke:
OK, fair enough! Let's close that MR! :)

PR #168, opened 2018-06-28
Add latest 10minutemail.* domains

ginkoid:
lakqs.com is from https://10minutemail.net.
kpooa.com is from https://10minutemail.org.
pqoia.com is from https://10minutemail.info.

They change the disposable domain every 45 days. These are the new ones.


martenson:
thank you @ginkoid

PR #169, opened 2018-07-09
Add active domains from tempr.email

pulzarraider:


ginkoid:
They all have a single MX record with varying priority, pointed at `mx.discard.email`. Seems good for inclusion.

martenson:
thanks @pulzarraider

PR #170, opened 2018-07-11
Updated PHP code snippet

Wruczek:
- Optimised and cleaned, faster by about 32.5% on my machine
- Added optional parameter to select the blacklist file, defaults to "disposable_email_blacklist.conf" file in the same directory as script
- Renamed and indented according to PSR-2
- Added notes about email validation and mbstring


martenson:
Awesome. Thanks @Wruczek !

PR #174, opened 2018-08-22
Add more domains

di:
[Zulip recently switched to using this library](https://github.com/zulip/zulip/commit/41f8618c04e9720f816ab0a5c8c50320928230ac), this adds a number of domains that were in their original list but are not in this list.

I've only added the domains with valid MX records, see below:

* `anotherdomaincyka.tk`
  * `anotherdomaincyka.tk. 17 IN MX 10 mail.b-io.co.`
* `antichef.com`
  * `antichef.com. 1626 IN MX 10 gourmet7.spamgourmet.com.`
* `antichef.net`
  * `antichef.net. 1517 IN MX 10 gourmet7.spamgourmet.com.`
* `arur01.tk`
  * `arur01.tk. 126 IN MX 10 mail.b-io.co.`
* `azazazatashkent.tk`
  * `azazazatashkent.tk. 265 IN MX 10 mail.b-io.co.`
* `bione.co`
  * `bione.co. 1026 IN MX 20 smx2.web-hosting.com.`
  * `bione.co. 1026 IN MX 30 smx3.web-hosting.com.`
  * `bione.co. 1026 IN MX 10 smx1.web-hosting.com.`
* `c2.hu`
  * `c2.hu. 127 IN MX 10 mx.c2.hu.`
* `chechnya.conf.work`
  * `chechnya.conf.work. 3599 IN MX 10 mailgw.ns36.de.`
* `ckaazaza.tk`
  * `ckaazaza.tk. 127 IN MX 10 mail.b-io.co.`
* `courriel.fr.nf`
  * `courriel.fr.nf. 1026 IN MX 1 courriel.fr.nf.`
* `dasdasdascyka.tk`
  * `dasdasdascyka.tk. 299 IN MX 10 mail.b-io.co.`
* `dfgh.net`
  * `dfgh.net. 3599 IN MX 10 gourmet7.spamgourmet.com.`
* `dm.w3internet.co.uk`
  * `dm.w3internet.co.uk. 10626 IN MX 10 mail.w3internet.co.uk.`
* `ficken.de`
  * `ficken.de. 184 IN MX 10 mailin100.dcpserver.de.`
* `flashbox.5july.org`
  * `flashbox.5july.org. 1799 IN MX 1 flashbox.5july.org.`
* `fragolina2.tk`
  * `fragolina2.tk. 14439 IN MX 10 mx.fragolina2.tk.`
* `frappina.tk`
  * `frappina.tk. 14439 IN MX 10 mx.frappina.tk.`
[...]

PR #172, opened 2018-07-30
add molms.com

almalki:
https://www.stopforumspam.com/domain/molms.com
https://github.com/ivolo/disposable-email-domains

PR #175, opened 2018-08-24
Add more CI checks

di:
Adds some more checks to the CI:
* checks that all domains are lowercase
* checks that domains don't appear more than once in a list
* checks that the lists are sorted alphanumerically
* checks that a domain doesn't appear in both the whitelist and blacklist

This also fixes some failures of those tests, namely that #172 added a domain that wasn't in the correct place alphabetically, and #174 added some domains that were in the whitelist already.


martenson:
perfect, thanks a bunch @di!

Issue #180, opened 2018-11-07
Disposable domains from temp-mail.org


mrpavlikov:
temp-mail.org provides mailboxes in different domains:

* 1thecity.biz
* acucre.com
* hiltonvr.com
* bnote.com
* nyrmusic.com
* oranek.com
* pachilly.com
* patonce.com
* rupayamail.com
* sandcars.net
* tryzoe.com
* zdenka.net

Got these ones by opening https://temp-mail.org/api/ in private tab multiple times, you can see the domain in the header. Probably should register app and try to get more with API, but these ones I got so far.

![image](https://user-images.githubusercontent.com/6112323/48140042-070e3980-e2b9-11e8-8990-76127ec4404e.png)

martenson:
@mrpavlikov Would you mind creating a pull request with these?

mrpavlikov:
@martenson got it done. Btw I had a bit of trouble with locale, Ubuntu 18.10 appears to have this one empty, so I had to set it manually, otherwise all the order was ruined. Maybe you should add this to your contributing guide.

```
export LC_ALL=C
```

PR #181, opened 2018-11-07
#180: added domains from temp-mail.org

mrpavlikov:
I've used these instructions on contributing:

```
$ cat disposable_email_blocklist.conf your_file | tr '[:upper:]' '[:lower:]' | sort -f | uniq -i > new_file.conf
$ comm -23 new_file.conf allowlist.conf > disposable_email_blocklist.conf
```

But second one said:

```
comm -23 new_file.conf allowlist.conf > disposable_email_blocklist.conf
comm: file 2 is not in sorted order
```

So I had to reorder it with this command first:

```
cat allowlist.conf | tr '[:upper:]' '[:lower:]' | sort -f | uniq -i > allowlist.conf
```

Still seems there is too many lines got reordered in both files, what am I doing wrong? I'm at Ubuntu 18.10.


mrpavlikov:
Guess it has something to do with locale

martenson:
Don't beat yourself over it, I will re-sort in the evening. Thanks for the
contribution!

Issue #149, opened 2018-03-06
Missing Domains from blacklist


C0nw0nk:
Here are some services who's domains are missing.

https://www.minuteinbox.com/

https://temp-mail.org/en/option/change/

http://tempsky.com/

https://www.moakt.com/

https://mailsucker.net/

https://dma.in-ulm.de/

simon21587:
nwytg.com

martenson:
Hi @C0nw0nk  Would you mind creating a pull request for these?

PR #185, opened 2018-11-09
Update disposable_email_blacklist.conf as disscused in #149

GermanS:
@martenson please, make review

### DIFF
--- a/disposable_email_blocklist.conf
+++ b/disposable_email_blocklist.conf
@@ -659,7 +659,6 @@ dingbone.com
 disaq.com
 disbox.net
 disbox.org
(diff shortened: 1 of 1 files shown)


martenson:
@GermanS what is the source of these domains?

Also, tests say:

```
The following domains appear twice in 'disposable_email_blocklist.conf':
* disbox.org
```

GermanS:
@martenson  those domains were stolen from #149

i.e. you could open url https://temp-mail.org/en/option/change/ and inspect domain field.

martenson:
Marvelous, thanks a bunch!

PR #191, opened 2019-01-18
gaggle.net is not an disposable domain

Tapac:
Please check https://www.gaggle.net/product/safetytipline/


martenson:
@Tapac thanks, and sorry if it affected your work

PR #193, opened 2019-01-21
Remove Sudo domains

rbholland:
Our company owns these Sudo domains.

We have made improvements to our app to limit the disposability of our email addresses (only paying users are allowed to use the app, we are working to limit the amount of email resets, etc.). Additionally, we have elaborate fraud monitoring, ensuring that our users are not spamming others. A significant amount of our users are reporting that their emails are getting blocked. We are working to clear our name from multiple blacklist providers, and noticed that several email providers rely on this repo.


martenson:
Thanks @rbholland !

PR #197, opened 2019-02-19
add emailbin.org

mtmail:
http://emailbin.org/

From HackerNews forum
```
emailbin.org is running on elasticsearch with a hideous frontend of pure raw html hahaha
usecases:
* email signups
* posting of email threads to government
* your ideas
it is and will always be a free place to lay your emails to rest :-) 
source code https://github.com/emailbin/web
```

### DIFF
--- a/disposable_email_blocklist.conf
+++ b/disposable_email_blocklist.conf
@@ -799,6 +799,7 @@ emailage.ga
 emailage.gq
 emailage.ml
 emailage.tk
+emailbin.org
 emaildienst.de
 emaildrop.io
 emailfake.ml


martenson:
Interesting service. Public _and_ searchable. Not sure what to make of it.

martenson:
what do you think @deguif ?

sloev:
creator of emailbin.org here. please ask if you have any questions @martenson

martenson:
@sloev is the emailbin service easy to use as a disposable email provider?

sloev:
emailbin.org is easy to use as a disposable email service but the intended use is unknown at this point. it is an experiment.
If you are afraid it will be used for signups etc then you can blacklist it.

My idea of use is:
```
i have a mail correspondance with the government about x
the correspondance is per se public information
i forward the correspondance to my_id@emailbin.org 
correspondance gets annonymized (emails removed) and indexed (with elastic search) and from there anybody can follow allong
```

martenson:
To me this seems so niche of a use case that unless it takes off as a disposable email service (at which point @sloev would probably stop being interested in running it) it seems overkill to blocklist it. Closing, but thanks for suggestion @mtmail and for participation @sloev.

PR #206, opened 2019-04-19
Added new domains from emailtemporal.org

cceglia:
Added new domains from emailtemporal.org and ordered them


martenson:
thanks @cceglia !

PR #218, opened 2019-09-28
Update allowlist.conf

AidanTheonas:
We are a tech company with the mission to accelerate access to clean, safer and affordable cooking solutions in developing countries using cutting edge software and hardware.

To learn more about KopaGas please visit https://www.kopagas.com/about


martenson:
I don't think this is a nice way to advertise yourself.

PR #228, opened 2020-01-13
Blacklist fosil.pro

Julien-Rcs:
cf https://temp-mail.us/


di:
Hi, could you also include the other domains from that provider:

* `rootfest.net`
*  `polyfaust.net`
* `cartelera.org`
* `hostcalls.com`

Thanks!

TheConceptBoy:
reptech.org

PR #232, opened 2020-03-13
Added additional temporary e-mail domain names

brtdv:
@martenson I've added a few additional temporary e-mail domain names to the list, cheers.

### DIFF
--- a/disposable_email_blocklist.conf
+++ b/disposable_email_blocklist.conf
@@ -29,6 +29,7 @@
 10minutemail.de
 10minutemail.ga
 10minutemail.gq
+10minutemail.it
 10minutemail.ml
 10minutemail.net
 10minutemail.nl
@@ -147,6 +148,7 @@
 99.com
 99cows.com
 99experts.com
+99pubblicita.com
 9mail.cf
 9me.site
 9mot.ru
@@ -317,6 +319,8 @@ anonymousness.com
 anotherdomaincyka.tk
 ansibleemail.com
 anthony-junkmail.com
+antichef.com
+antichef.net
 antireg.com
 antireg.ru
 antispam.de
@@ -390,6 +394,7 @@ barryogorman.com
 bartdevos.be
 basscode.org
 bauwerke-online.com
+baxomale.ht.cx
 bazaaboom.com
 bbbbyyzz.info
 bbhost.us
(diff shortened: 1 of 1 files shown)


disposable_email_blocklist.conf
```diff
@@ -799,6 +809,7 @@ ecallheandi.com
 ecolo-online.fr
 edgex.ru
 edinburgh-airporthotels.com
+edu.sg
```
di: This is not a domain, it's a public suffix, which is causing CI to fail here.

di:
Hi @brtdv, thanks for the contribution. Can you provide a source for these domains? E.g., where can I go to get a `@antichef.com` temporary email address?

Thanks!

brtdv:
@di Good question, I've tried looking for the source by looking back in the commit history of my web app, which contains a similar list of domain names that are blocked.

I was looking to update my list in my webapp, since the latest commit to my file was from june '18. That's how I found [martenson's repo](https://github.com/martenson/disposable-email-domains), which I then used to update the list of 'blocked' e-mail domains in my web app. I thought I'd contribute back by adding all missing domains that were in my list, which weren't in the repo...

So that's the reason behind my commit... Unfortunately I'm not able to provide a real 'source'. You can. however, find the domain name on [check-mail.org](https://check-mail.org/domain/antichef.com/) and verify that this is indeed a temporary e-mail domain name, that can be created by using [spamgourmet.com](https://www.spamgourmet.com/index.pl).

disposable_email_blocklist.conf
brtdv: You are probably correct, this [does not seem like a disposable e-mail address](https://check-mail.org/domain/edu.sg/). My bad, should I add an extra commit removing this item?

DanielAbsi:
Hi I found new temporary emai to be added: 
mailnd7.com
qortu.com
psk3n.com
tashjw.com
ifavorsprt.com
andyes.net

disposable_email_blocklist.conf
di: Yes, please remove it. Thanks!

di:
Thanks @brtdv. Without a source it's hard to verify these, e.g. https://check-mail.org/domain/10minutemail.it/ says that this is not a disposable domain (and doesn't even have an MX record).

I'd be OK with merging any domains that a) have an MX record, and b) that MX record points to a domain already on the list.

martenson:
Since this PR has stalled I am closing it. Please feel free to reopen when you can provide source for added domains.

Thanks for contributing.

PR #241, opened 2020-06-19
add more domains

willtho89:
These domains were used to try and cheat on several of our platforms where users can vote on things. 
The list contains entries we/our system determined as spam/disposable mails and was started somewhere around ~2015. Unfortunately I can not vouch for the validity of these domains, but we heard no complains of actual users.


willtho89:
Nevermind, we have Domains which might not be seen as disposable (e.g. aol.de)
Might do a cleanup and make a new PR eventually

PR #242, opened 2020-06-19
Add domains from temporary-mail.net

Tapac:
Related to https://github.com/martenson/disposable-email-domains/pull/240 but without allowlist changes.


di:
Thanks!

PR #247, opened 2020-07-20
Add maskedmails.com disposable domain

Tapac:
Link to check: https://check-mail.org/domain/maskedmails.com/


di:
Hi @Tapac, thanks for the contribution. Do you know where one could generate an `@maskedmails.com` email address?

martenson:
@Tapac unless you can provide a reason, why this domain should be considered disposable, we cannot add it. Please get back to us with any new info. Thanks!

PR #255, opened 2020-12-13
Update disposable domains

bbkgh:
According to is-disposable-email-domain npm package


martenson:
Hi @bbkgh, I think we don't want to do this. Our project has a goal to provide a verified and maintained list. Throwing in a dump of 100k entries from some new list which is based on a [single commit](https://github.com/Igor-Rabodzei/isDisposable/commit/fcf193c32c71c8d86676a1357601f0cfe4b21449) with no explanation or sources defeats that. We need to have at least some means to verify sources and reason about data.

bbkgh:
> Hi @bbkgh, I think we don't want to do this. Our project has a goal to provide a verified and maintained list. Throwing in a dump of 100k entries from some new list which is based on a [single commit](https://github.com/Igor-Rabodzei/isDisposable/commit/fcf193c32c71c8d86676a1357601f0cfe4b21449) with no explanation or sources defeats that. We need to have at least some means to verify sources and reason about data.

You are right. I saw a disposable email domain today which was not in your list then i found that library and it is more complete (not sure about correctness).
However i am sorry for this improper and unprofessional pull request (and for not reading contributing section in readme) But i think this strict policy (and manual checking) can make this library outdated and less useful for this special context (but it's up to you).
Also thanks for your polite and kind reply.

martenson:
@bbkgh Projects such as ours consist of "fighting the windmills" because creating a new domain and enlisting it as disposable is trivial, dirt cheap, and infinite. Furthermore domain names expire, change owners, change purpose etc. We are fully aware of the above and aim to provide a layer of curation instead of taking a mortar canon and shooting down everything that moves.

We'd welcome a PR with the disposable service you encountered. Thank you for considering how to help this project.

PR #261, opened 2021-02-21
Added a Russian disposable domain

karolyi:


karolyi:
bump

karolyi:
I can haz merge plox?

martenson:
@karolyi Hi. How can I obtain a disposable address with this domain?

karolyi:
I have no idea. Last time I checked there was a "please write an email address to register' on the site, My projects got random emails registered from this domain, as if they were bots. If you search for the domain, you'll see that other people had the same issue.

martenson:
Thanks for the PR @karolyi but without you sharing a way how to create disposable addresses with these, we cannot merge. Feel free to reopen with more info.

karolyi:
@martenson do you really expect me to write to the address listed on their site to get a disposable email address they use just to prove to you that it *is* a disposable domain?

you could have just followed my advice and do some web searches to see why I added this PR. you can also see that I am the author of the py3-validate-email package that uses this list.

I don't commit new stuff on here just out of fun.

martenson:
@karolyi I understand how the service is supposedly working and I am not asking you to send them an email nor am I accusing you of making this commit out of fun.

It is unclear to me whether this qualifies as "disposable" for the purposes of this list since you have to register using a functioning email address and you need an email client which put together is a relatively high barrier of entry. 

I did test sending email to the `register@thefmail.com` address about 30m ago and received no reply, so it seems no real automation of disposable addresses is in place.

I am open to discussing this though. What do you think @di @deguif ?

jarthod:
Just for the record (I understand if this is not enough proof), the https://www.ipqualityscore.com/domain-reputation/thefmail.com website I discovered when doing #271 also considers `thefmail.com` as disposable.

martenson:
I have not been able to get a disposable address with this service.

PR #263, opened 2021-02-25
Add a shell script for maintaining disposable_email_blocklist.conf

HollowMan6:
My system's localization settings are Chinese, I won't produce the same order as the repo list using `cat disposable_email_blocklist.conf your_file | tr '[:upper:]' '[:lower:]' | sort -f | uniq -i > new_file.conf` unless executing `LC_ALL=C; export LC_ALL` beforehand.

image

Signed-off-by: Hollow Man <hollowman@hollowman.ml>

### DIFF
--- a/README.md
+++ b/README.md
@@ -113,9 +113,7 @@ Feel free to create PR with additions or request removal of some domain (with re
 
 Specifically, if adding more than one new domain, please cite in your PR where one can generate a disposable email address which uses that domain, so the maintainers can verify it.
 
-Use:
-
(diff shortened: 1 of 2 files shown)


di:
Thanks! We probably should just add a shell script that does this command instead, and sets the local. Would you be able your PR to add that instead?

HollowMan6:
> Thanks! We probably should just add a shell script that does this command instead, and sets the locale. Would you be able your PR to add that instead?

@di Done!

README.md
```diff
@@ -115,11 +115,7 @@ Specifically, if adding more than one new domain, please cite in your PR where o
 
 Use:
```
di, line 117: I think these lines need removed as well.

HollowMan6: Done! Sorry for the redundancy.

di (approved):

di:
Thanks!

martenson:
Excellent, thanks @HollowMan6 !

PR #264, opened 2021-02-26
add disposable email domains found in the wild

sansculotte:
botfed.com, eamarian.com, ezeca.com, mailpoof.com, maksap.com, overcomeoj.com, pashter.com, qakexpected.com, safemail.icu, sixdrops.org

### DIFF
--- a/disposable_email_blocklist.conf
+++ b/disposable_email_blocklist.conf
@@ -453,6 +453,7 @@ borged.com
 borged.net
 borged.org
 bot.nu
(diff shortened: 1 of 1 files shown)


di:
Can you show us how to create disposable emails with these domains, or where these came from?

martenson:
Thanks for the contribution @sansculotte but without you sharing a way how to create disposable addresses with these, we cannot merge. Feel free to reopen with more info.

PR #265, opened 2021-03-01
Added crawled domains from 'generator.email'

peterfroehlich:
We found a lot of people started using [generator.email](https://generator.email) and their verticals. 
I used a simple python script to crawl over the domain search:

https://gist.github.com/peterfroehlich/5590f1dfbad41b0754efbf861555acd2

Result after cleaning out garbage and subdomains: over 9100 domains.


martenson:
Whoa, this is a giant list @peterfroehlich ! The page which guides users to share password for their domain registration also seems pretty scammy (https://generator.email/blog/new-email-generator-domain). 

What do you guys think of this approach @di @deguif ?

jarthod:
Hello! That's a big list indeed, but I noticed most of them don't have any MX record so we could probably skim it down further by removing all these broken domains? (of course it's possible some of these will get an MX some day but I guess it's more likely they will never). If needed I can have a look at a script to do that.

By the way maybe this could be interesting for the existing list too? maybe we could add an (optional?) MX test to `verify.py` ?

martenson:
@jarthod those are both very good ideas I think. Contributions in this direction will be most welcome.

peterfroehlich:
I added a separate verification script for the records. 
Separate because it takes quite some time to iterate over all the records (without multithreading and potentially run into rate limiting) and the scope seems different (linting and sanity check vs e2e test (kinda)).

I'm scraping the list again and will filter the broken records.

tompec:
Just a heads up, a domain can still receive emails without MX records. Mail servers will fall back to the A record.

martenson:
@tompec indeed, that makes things complicated [source](https://tools.ietf.org/html/rfc5321#section-5.1)

> If an empty list of MXs is returned,
   the address is treated as if it was associated with an implicit MX
   RR, with a preference of 0, pointing to that host.

peterfroehlich:
> Just a heads up, a domain can still receive emails without MX records. Mail servers will fall back to the A record.

Darn ^_^
Well the script is still filtering out invalid/unreachable domains, maybe it still helps in some way. 

Regarding the mail domains from generator.email, their interface seems to flag domains without MX record as "not supported" so it makes sense to drop them from the scraped list.

jarthod:
> Just a heads up, a domain can still receive emails without MX records. Mail servers will fall back to the A record.

Oh, I had no idea :) that's interesting. Indeed for this list if generator.email considers them invalid with no MX I guess it still makes sense to filter them out to keep the list shorter. And then for the rest maybe we should only remove domains if both MX and A are missing? and if there's not too many in this situation (missing MX but A record present), we can probably check them to see if the server in A is actually a mail server? (ports open). Probably not worth implementing that just to clean a couple domains off the list.

peterfroehlich:
I've pulled the whole list again from generator.email and filtered out invalid records / records without MX. 
Then merged them with the current "master" list. 

The gist in the first post is updated with the new logic.

zhangjiayin:
we found  something like  generator.email should  contain 

 generator.email
  emailfake.com
  tempm.com
  mx2.titan.email
  mx1.titan.email
  mx4.mxstorm.com
  mx4.mxcomet.com

stuartnelson3:
What's the status of this? Would love to get all these new domains

martenson:
active `generator email` domains got merged in https://github.com/disposable-email-domains/disposable-email-domains/pull/903

If you'd like to contribute the dns check script please open a new PR. I am happy to include such tool in some form.

edit: @peterfroehlich I extracted your DNS check to https://github.com/disposable-email-domains/disposable-email-domains/pull/904

PR #268, opened 2021-04-20
Additional temporary domains

gdarko:
The list that I want to merge can be found on https://tempmail.plus 

Spammers actively use this service to register free accounts on our sites


martenson:
Thanks a lot @gdarko !

PR #272, opened 2021-05-20
Add supersave.net

hkonsti:
Hey :wave:,
this email service was causing me a lot of trouble with tons of throw away emails. I thought it would be useful to add it.
You can visit their website [here](http://supersave.net).

By the way, thanks for this great list!


hkonsti:
Emails can be created and accessed using this website: https://mail.tm/en/

martenson:
splendid, thanks @hkonsti !

PR #276, opened 2021-06-13
add instasmail.com, yousmail.com, smailpro.com, hatmail.ir

payam124:
instasmail.com and  yousmail.com are both from smailpro.com

### DIFF
--- a/disposable_email_blocklist.conf
+++ b/disposable_email_blocklist.conf
@@ -1289,6 +1289,7 @@ hartbot.de
 hasanmail.ml
 hat-geld.de
 hatespam.org
+hatmail.ir
 hawrong.com
 haydoo.com
 hazelnut4u.com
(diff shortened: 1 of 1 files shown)


martenson:
This one is a pickle, since smailpro.com seems to automate gmail registration as a disposable email service (I got addresses like `clinicpainfu.l36.8.3@gmail.com`. This is surely abuse of google service and should be reported upstream. 

What do @di and @deguif think?

di:
Does it actually work? I wasn't able to get it to receive a message to a `@gmail.com` address it generated.

martenson:
I tested it with guerillamail and received the message fine :/

Advanced

martenson:
This might be a form to use for reporting gmail abuse:
https://support.google.com/mail/contact/abuse

Mis012:
can confirm that smailpro.com works nicely to avoid stupid login walls, thanks for the tip :)

martenson:
@Mis012 just be careful about who reads your mail ;)

PR #277, opened 2021-07-01
Added webhook.site

io238:
This is a web service used to debug webhooks and emails. Generates disposable email addresses under the `email.webhook.site` domain.

Details: https://docs.webhook.site/index.html


di:
Thanks! We only add primary domains to this list, not subdomains, so this can just be `webhook.site`

di:
Thanks!

PR #278, opened 2021-07-26
Adds bptfp.net

anotherstarburst:
Adds bptfp.net an alias of https://10minutemail.com/


martenson:
Thank you @anotherstarburst !

PR #279, opened 2021-08-09
Added new domains iiron.us, ttht.us

arlek777:


martenson:
Closing since a way how to test these was not provided.

PR #282, opened 2021-09-10
add uba3.be

nhosoya:
Add uma3[.]be

Details: 
https://m.kuku.lu/
https://m.kuku.lu/config_domain.php


nhosoya:
I've included the same change in https://github.com/disposable-email-domains/disposable-email-domains/pull/284 and close it.

PR #285, opened 2021-09-28
Make maintain.sh executable

nhosoya:
This is a trivial change.
I think it is more convenient if maintain.sh is executable.

```sh
chmox +x maintain.sh
```


martenson:
Seems useful, thanks @nhosoya !

PR #293, opened 2021-12-06
Add domains available on https://mail.td/

nhosoya:
Add the following domains to the block list

- getover[.]de
- tdtda[.]com
- uuf[.]me

Email addresses for these domains can be obtained from https://mail.td


nhosoya:
@martenson @di
Thanks for all the maintenance.
Please give me a review.

nhosoya:
thanks

PR #295, opened 2021-12-13
add moakt emails

X-SLAYER:
Add emai laddresses from this domain https://www.moakt.com/

### DIFF
--- a/allowlist.conf
+++ b/allowlist.conf
@@ -170,3 +170,14 @@ yahoo.com.vn
 yeah.net
 yepmail.net
 your-mail.com
+tmpbox.net
+moakt.cc
+disbox.net
+tmpmail.org
+tmpmail.net
+tmails.net
+disbox.org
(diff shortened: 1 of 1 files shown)


allowlist.conf
```diff
@@ -170,3 +170,14 @@ yahoo.com.vn
...
+moakt.cc
+disbox.net
+tmpmail.org
+tmpmail.net
+tmails.net
+disbox.org
+moakt.co
+moakt.ws
+tmail.ws
+bareed.ws
```
nhosoya, line 183: Looking at [this site](https://www.moakt.com/), it appears that these domains are disposable.
If so, they should be added to `disposable_email_blocklist.conf`, not `allowlist.conf`.

di: Good catch!

martenson:
All these are actually already in. Thanks @X-SLAYER

PR #297, opened 2022-01-11
add mjj.edu.ge and lsyx24.com

ym:


disposable_email_blocklist.conf
```diff
@@ -1914,6 +1915,7 @@ mirai.re
 misterpinball.de
 miucce.com
 mji.ro
+mjj.edu.ge
```
di, line 1918: How can I generate a disposable email from this domain?

ym: https://mail.mjj.edu.ge/

PR #244, opened 2020-06-28
Update disposable_email_blocklist.conf

tuptuptuu:
add hey.com
DA0E22A7-460A-432A-8D9D-12AA53F8577B

The [iOS app](https://apps.apple.com/gb/app/hey-email/id1506603805) allows you to create free disposable email addresses.

Issue #46, opened 2016-07-06
decide on what domain levels to cover


martenson:
In https://github.com/martenson/disposable-email-domains/pull/45 we discussed whether we should only gather domains in xxx.yy form or whether to also accept third (and deeper) levels such as zzz.xxx.yyy

This might need some research as there are cases that are clearly exempt (.co.uk) and some appear unclear. Also if we remove zzz.xxx.yyy should we always keep xxx.yyy as this will vastly broaden the list impact? Is the assumption that when spammer controls xxx.yyy they also control zzz.xxx.yyy reasonable?

natefoo:
I think this only makes sense in cases where the "TLD" is xxx.yyy instead of yyy. Otherwise, if a spammer controls the second level domain, they control everything underneath. The exception would be service providers (e.g. dyndns, but no legitimate mail should come from there, I just use it as an examnple).

PR #298, opened 2022-01-14
Add email -> relay.firefox.com

Mathieu-5:
Add domain name from https://relay.firefox.com/

### DIFF
--- a/disposable_email_blocklist.conf
+++ b/disposable_email_blocklist.conf
@@ -3272,3 +3272,4 @@ zymuying.com
 zzi.us
 zzrgg.com
 zzz.com
+relay.firefox.com


di:
@Mathieu-5 FYI, linting is failing, the list needs to remain alphabetized.

di (commented):
I think I'm in favor of this. This is similar to #244, but slightly different in that the sole purpose of this domain is to generate disposable emails.

martenson:
Well, nice pickle. Why are you doing this to us Firefox? Among other things this throws a wrench in the original (not really rock-solid) reasoning about domain levels from [here](https://github.com/disposable-email-domains/disposable-email-domains/issues/46) -- so it breaks our CI even if in the correct order.

Looking at the screenshots [here](https://addons.mozilla.org/en-US/firefox/addon/private-relay/) isn't the domain actually supposed to be `mozmail.com`? That one I'd be ok with including I think.

di:
Looks like you're right, it's either `@mozmail.com`, `@relay.mozmail.com` or `@personaldomain.mozmail.com`.

My reasoning on including this is that an email with a mozmail domain is never going to be a primary email and is always going to forward to some other address.

worldofgeese:
Your GitHub repo looks like it's used, or can potentially be used, as a weapon by providers trying to rob users of one of the few defenses they have to their email address leaking, a scarily common occurrence, and is then weaponized by bad actors to flood those users inboxes with spam.

Can you not do this? You look like extremely bad actors. Please don't contribute to an unsafe internet.

I use Private Relay to protect my personal mail address, not as a tool for spam. I'm not even sure how a user would use Private Relay for spam, as users cannot begin email chains with a Relay address, only respond to mails delivered to those addresses.

codemicro:
Additionally, these forwarding addresses don't expire or go away until a user explicitly deletes them - they're purely a privacy tool. It's roughly equivalent to someone registering a random, single-use Google email address for a single service.

digitalblossom:
I can understand the concept behind this repo, assisting in preventing abuse of free offers etc.

However I also agree with my previous speakers in that email aliases are the only tool to prevent personal mail adresses from turning up in databreaches and being spammed.

So please don't do this, it will most likely do more damage than good. Developers of services that are being abused with temporary mails can still ban actors manually no?

ColourGrey:
Well, congratulations! You've made it big on [Hacker News](https://news.ycombinator.com/item?id=29959076), you're famous now! As in a famous song, from now on you'll get ["money for nothing and chicks for free"](https://www.youtube.com/watch?v=wTP2RUD_cL0)!

Joking aside, could you please reconsider the decision to block these Mozilla-related domains? Right now, the proposal got 10 downvotes in less than 5 minutes, which should clearly convey the geek-ish public's opinion on the matter. Just look at the ratio between upvotes and downvotes of the first post in this topic and you'll understand what you've triggered. Thank you.

msoopern:
You're a fucking asshole.

martenson:
xref: https://github.com/wesbos/burner-email-providers/pull/339

martenson:
I'd like to highlight a post here, that imho does a good job explaining some of the reasoning behind this repo: https://news.ycombinator.com/item?id=29959744

martenson:
Mozilla Relay requires proper registration, offers 5 "masks" for free after that, and requires $ 1 monthly subscription for more.

Looking at this I do not believe this passes as "disposable". I am going to close this PR now. Thank you for opening this issue @Mathieu-5.

groovecoder:
Thanks @martenson !

89z:
https://github.com/CodeKJ/DISIFY/issues/25

PR #306, opened 2022-02-07
feat: adding new 10minutemail domains

russorat:
adding new domains from https://10minutemail.net/history.html


martenson:
Thanks @russorat; much appreciated.

PR #312, opened 2022-02-25
Add reimondo.com from temp-mail.org

renjithspace:


russorat:
uses the same ip as https://temp-mail.org/en/ - 174.138.127.74

image

Added it to https://github.com/disposable-email-domains/disposable-email-domains/pull/314 and ran `./maintain.sh` script to keep it in order.

boywithkeyboard:
@martenson this pull request can be closed. `reimondo.com` is already in the list at [line 2522](https://github.com/disposable-email-domains/disposable-email-domains/blob/1ac927fbbadf31f4893bada59c8d14688dc3d62e/disposable_email_blocklist.conf#L2522).

PR #313, opened 2022-02-28

edgarasben:
Add ishop2k.com

Generated by https://temp-mail.org/


martenson:
@edgarasben I think @russorat means it as an indicator that the domain is indeed part of a disposable pool

PR #323, opened 2022-04-20
Adds PHP Composer package details to README

elliotjreed:
It appears a few people are using this package already, so here it is in the README should others see it here first and think it may be of use. It's updated daily in GitHub actions and tagged as a patch release for Composer updates on Packagist. Suggestions welcome!


martenson:
Thanks @elliotjreed !

PR #324, opened 2022-04-25
Add domain from https://www.linshi-email.com/en

m3nu:
China-focused disposable email service: https://www.linshi-email.com/en

Screen Shot 2022-04-25 at 13 52 13


martenson:
thank you @m3nu

PR #326, opened 2022-05-17
Add new domain from temp-mail.org

silas-riacourt-km:
Add new domain from https://temp-mail.org/fr/ > @cupbest.com
image


martenson:
thanks @silas-riacourt-km

Issue #273, opened 2021-05-20
fight the disposable automation with an anti-automation 


martenson:
If the https://web1.temp-mail.org/request/domains/format list is something that is fairly stable we could implement a CI check that automatically merges domains in that list with our blocklist.

```xml
<xml>
<item>@64ge.com</item>
<item>@brayy.com</item>
<item>@frnla.com</item>
<item>@geekale.com</item>
</xml>
```

SufiSR:
> If the https://web1.temp-mail.org/request/domains/format list is something that is fairly stable we could implement a CI check that automatically merges domains in that list with our blocklist.
> 
> ```
> <xml>
> <item>@64ge.com</item>
> <item>@brayy.com</item>
> <item>@frnla.com</item>
> <item>@geekale.com</item>
> </xml>
> ```

The endpoint is stable but tends to outdate quickly. It also lists only four disposable mails although there are more than just those four.

martenson:
@SufiSR I thought of something like a daily CI run - those listed there seem to be live which makes them valuable to us. Maybe we could look for more endpoint like this.

DanielAbsi:
Hey @martenson,
Do you also have the URL for, emailondeck.com domain to be blocked like you did with temp-mail.com?

DanielAbsi:
Link is down for https://web1.temp-mail.org/request/domains/format

martenson:
elaborated the issue in https://github.com/disposable-email-domains/disposable-email-domains/issues/450

PR #327, opened 2022-06-17
Add domains used by 10minutemail.net

alexmv:


alexmv:
They [brag that they change the domains every 45 days](https://10minutemail.net/about.html), so this may not be very effective in the game of whack-a-mole.

martenson:
Ideally we would automate the whack-a-mole, something like was outlined here https://github.com/disposable-email-domains/disposable-email-domains/issues/273

martenson:
Thanks for the contribution @alexmv! Side note: it has been more than 45 days since the domains you've added have been used so the rotation won't be as hot as they sell it.

PR #329, opened 2022-08-02
Improvements for the Python example code

Holzhaus:
First of all this fixes the issue the `open()` call is missing. Otherwise `blocklist` is a string instead of a file, and doesn't have a `readlines()` method.

I also added some small improvements. I know it's just a simple code sample, but most people will probably copy&paste it and use it as a base instead of writing new code from scratch. Hence, I made added some small improvments that improve the code without making it much longer or more complex. These changes are:

- Use an unordered `set` instead of an ordered `list` to reduce the lookup complexity from *O(n)* to *O(1)*.
- Use a `with` statement to close the blocklist file automatically after reading
- Use `str.partition` instead of `str.split` for more consistent results in case `email` is not a valid email address.

See the individual commit messages for details.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,7 +16,7 @@ Example Usage
 **Python**
 ```Python
 with open('disposable_email_blocklist.conf') as blocklist:
-    blocklist_content = [line.rstrip() for line in blocklist.readlines()]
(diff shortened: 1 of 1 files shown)


martenson:
thanks @Holzhaus !

PR #330, opened 2022-08-04
add https://tempmail.plus domains

mpstenson:
The domains can be found here https://tempmail.plus/en/#!


martenson:
thank you @mpstenson !

PR #337, opened 2022-09-08
feat: add domains from https://emailfake.com

mlgr-io:
I've added 16 new domains from https://emailfake.com (that domain itself plus 15 domains you can choose from at the top of that page).


martenson:
Thank you @mlgr-io !

PR #341, opened 2022-10-26
Add cdfaq com and nezid.com domains

m-keil:
Hey, just catched a few domains, which were not part of the list at https://10minutemail.net/ while testing my implementation of the list.
See attached files:

10minutemail-cdfaq

10minutemail-nezid

Cheers,
Max


martenson:
thanks @m-keil !

PR #342, opened 2022-11-14
Include a few extra temporary domains

derrix060:
I got those emails from https://tempmailo.com/


martenson:
thanks @derrix060, please run `maintain.sh` on your PR

PR #344, opened 2022-11-16
Add fallinhay.com from https://www.fakemail.net/

mpstenson:
Add fallinhay.com from https://www.fakemail.net/

example screenshot:
Screen Shot 2022-11-16 at 9 57 16 AM


martenson:
thanks @mpstenson!

PR #348, opened 2022-12-10
Added additional disposable domain

sdoering:
Added tmmbt.net as domain. I just had a registration from this domain and checked it as a disposable 10 minute email provider.


martenson:
thanks @sdoering !

PR #350, opened 2022-12-13
Added Bash script example, + spymail.com domain

shamo0:
Added Bash script example, + @spymail.com domain. email can be generated at spymail.com


martenson:
thanks @shamo0 !

shamo0:
My pleasure 👍🏻

PR #351, opened 2022-12-26
Add emails to list

d-knafo:
Emails addresses can be generated at https://temp-mail.org/  and at https://www.disposablemail.com/


martenson:
thanks @d-knafo ! Could you please reorder the domains according to the format we use? (or just run maintain.sh)

PR #359, opened 2023-03-14
add more domains

stefansaraev:
:)


martenson:
Hi @stefansaraev -- what are the sources of these domains and how can I obtain a disposable email with them? I checked some from your list and they were often domains for sale or unregistered.

PR #362, opened 2023-04-16
Missing some domains.

oracularhades:
note: .fr.cr was being used here: https://yopmail.com/en/alternate-domains

other link: https://www.disposablemail.com/


oracularhades (commented):
removed double

oracularhades:
errrr, that action looks like it messed up?

martenson:
I don't think it did

> * 'doojazz.com' should come before 'doquier.tk'

PR #365, opened 2023-05-04
Add fextemp.com to blocklist

patrick-clearbooks:
Addresses `@fextemp.com` can be registered at https://tempmail.plus


martenson:
thanks @patrick-clearbooks !

PR #369, opened 2023-05-06
Add developermail.com

JodanJodan:
[developermail.com](https://developermail.com/) -> `developermail.com`


martenson:
thanks @JodanJodan

PR #371, opened 2023-05-08

hemanthmantri:
Added email domains that could be generated using tempmailo.com


boywithkeyboard:
@hemanthmantri run the maintain script (`./maintain.sh`) to add these domains in the correct order!

PR #387, opened 2023-06-25
add two new domains from 1secmail.com

stuartnelson3:
Two new domains listed at 1secmail.com:

image


martenson:
thanks @stuartnelson3 !

PR #391, opened 2023-07-07
Add some domains from InstAddr

grj1234:
The disposable e-mail domains from InstAddr ( https://m.kuku.lu/ )
"instmail.uk" domain is used for the disposable e-mail address with expiration.
"eripo.net", "nyasan.com", and "shchiba.uk" domains are available only in Japan.


martenson:
thanks @grj1234 !

PR #395, opened 2023-07-23
Add ridteam.com to blocklist

apodolny:
This domain is used by https://temp-mail.org/


nillpoe:
@martenson I think this PR should be closed as **irreproducible**.

martenson:
thanks for your suggestion but we want to verify the domain is actually used by a disposable provider by ourselves. We need a screenshot or similar.

PR #396, opened 2023-07-25
Add sparkroi.com to disposable_email_blocklist.conf

Sabir001:
ZeroBounce identifies this as disposable email domain.  We had some issues with this domain. And then checked in ZeroBounce and found that this is a disposable email domain.


nillpoe:

PR #402, opened 2023-08-10
Add 7 new domains from InstAddr

grj1234:
The disposable e-mail domains from InstAddr ( https://m.kuku.lu/ )


martenson:
thank you @grj1234 !

PR #403, opened 2023-08-25
Update webhook.site to email.webhook.site

luisfontes19:
Emails from https://webhook.site/ are `email.webhook.site` instead of just `webhook.site`


luisfontes19:
Closing this as I noticed that the lis is only the top domain

PR #404, opened 2023-08-27
Add 4 new domains from InstAddr

grj1234:
The disposable e-mail domains from [InstAddr](https://m.kuku.lu/)
"onlyapp.net" domain is available only in Japan.
"akugu.com" and "pwpwa.com" domains are originally available for premium users, but it seems available also for free users now.
Screenshot


martenson:
thanks @grj1234 !

PR #405, opened 2023-09-05
Add domains from muellmail.com.

SteffenLanger:
> If you are adding domains please state where one can generate a disposable email address which uses the added domains, so the maintainers can verify it. Without this information your PR will be closed.

You can create those domains at https://muellmail.com.

image


martenson:
thanks @SteffenLanger !

PR #406, opened 2023-09-07
Addition of mailers.edu.pl , outlook.edu.pl and privmail.edu.pl from https://etempmail.com/zh

Txori:
Addition of mailers.edu.pl , outlook.edu.pl and privmail.edu.pl from https://etempmail.com/zh


martenson:
thank you @Txori !

PR #415, opened 2023-10-11
Add "docasnyemail.sk"

disposable-updater:
These domain(s) is(/are) used for docasnyemail.sk


disposable-updater:
Sorry about that I mistakenly made a pull request for other commits at the same time 🙏🏻

PR #420, opened 2023-10-24
Added serveral domains to block list

jonathon-love:
these domains are provided here: https://10minutemail.com/

if you run the following bash script, it will generate a selection for you:

```
for run in {1..10}; do
  curl 'https://10minutemail.com/session/address' -H 'User-Agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10.15; rv:109.0) Gecko/20100101 Firefox/119.0';echo '\n'
done
```

although if you do it too much, it does start returning:

```
{"timestamp":"2023-10-24T07:20:51.225+00:00","status":403,"error":"Forbidden","path":"/session/address"}
```

so i assume it blacklists IPs, but i'm not sure.

### DIFF
--- a/disposable_email_blocklist.conf
+++ b/disposable_email_blocklist.conf
@@ -553,6 +553,10 @@ byom.de
 c51vsgq.com
 cachedot.net
 californiafitnessdeals.com
+cazlg.com
+cazlp.com
+cazlq.com
+cazlv.com
 cam4you.cc
 camping-grill.info
 candymail.de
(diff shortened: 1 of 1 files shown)


nillpoe:
bump

martenson:
10minutemail is behind captcha so curling won't work; we need to be smarter to automate

jonathon-love:
oh well, in october 2023 when i opened this, you could curl it no problems :P

martenson:
@jonathon-love I do believe you, but the providers are getting more elaborate. They're often behind cloudflare, rotate domains daily etc. The domains you mentioned are likely not used anymore and I apologize for not getting to you PR sooner.

PR #426, opened 2023-11-24

guiguille:
On Yopmail.com, a full list of alternative domains can be found here :
https://yopmail.com/fr/alternate-domains


martenson:
we're scraping these automatically now (https://github.com/disposable-email-domains/disposable-email-domains/blob/main/fetch_domains.py#L58), thanks @guiguille

PR #435, opened 2023-12-15

kachkolasa:
I found two more temp emails that were not in this list, just added them!

### DIFF
--- a/disposable_email_blocklist.conf
+++ b/disposable_email_blocklist.conf
@@ -737,6 +737,7 @@ deadspam.com
 deagot.com
 dealja.com
 dealrek.com
+deecie.com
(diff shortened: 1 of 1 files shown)


nillpoe:
image

the first one, `deecie.com` has been already suspended.

image

and the second one, `wireconnected.com` claims that they're not throwaway email service, and they're legit.

image

and their website seems pretty legit to me.

nillpoe:
@martenson I think this PR should be closed as **outdated** & **irreproducible**.

otherwise:
@kachkolasa please provide strict proof that `wireconnected.com` and `smtp.dev` are actually disposable email providers.

martenson:

Issue #84, opened 2017-03-10
False positives


szepeviktor:
`< disposable_email_blacklist.conf xargs -n 1 host -t MX` reveals that

- there are domain with such MX-s as `aspmx.l.google.com.`  (http://www.drevo.si/)
- or parked domains (NS is ns1.dnparking.com.)
- or have a new registrant  (http://www.emil.com/)
- or expired
- or having no Mx at all

Please advise.

szepeviktor:
One solution would be to use Travis CI:

1. Is it a confirmed disposable email domain?
1. Is it registered? - whois
1. Does it have an NS? - root servers
1. Does it have an MX - DNS
1. Are **all** of the MX-s on our MX whitelist?

martenson:
These are great ideas. I am not that educated in domain system - with no MX record can the domain be used for mailing in any way?

szepeviktor:
> with no MX record can the domain be used for mailing in any way?

Of course not.

martenson:
@szepeviktor So I guess good first step would be to move domains with no MX record out of the list. Something like `discontinued_domains.conf`.

chafreaky:
Mhmm, what about people who use this list to prune databases from disposable email addresses, both old and current? Most disposable email providers have a 45 days sunset policy and the domains are then dropped after that time. Should they not be listed anymore?

martenson:
@chafreaky That is a use case that would be covered by transferring the non-existent-MX domains to `discontinued_domains.conf` - would it not?

szepeviktor:
This issue is more about some kind of quality control, keeping thing up-to-date.

martenson:
@szepeviktor To be frank I think non-existent MX entry does not make (former) disposable domain trustworthy per se.

chafreaky:
Definitely not. And a discontinued file would work. We'd have to update the
readme to mention the possibility of including both lists when pruning.

ereli:
Have a look at https://github.com/ereli/disposable-email-domains/tree/validate 
you can run python  `./validate_dns.py |awk -F" - " '{print $2 }'|sort|uniq -c` 
and find out how many domains are no longer actively used.  I should  probably retry on NoAnswer, but these are the basic numbers I've got: 
```
1355  MX found
49  no resolver NS servers found
384  non existent domain
391 NoAnswer
```

ScottyTM:
Domains without an MX can still receive mails. The MTA will try the A record to deliver mails to this domain.

[...]

Issue #58, opened 2016-08-27
Clean-up dead domains


marcelstoer:
I suggest you run an automated clean-up task from time to time that removes dead domains from your collection. You won't find domains like http://ww8.10minutemail.de/ or http://1mail.ml/ which still exist but may not be used for disposable mail addresses anymore but you'd find domains like http://30minutemail.com/ for which there's no DNS entry anymore.

irvin:
I think we should move dead domain into a separate list, because some people will still use those dead domain email address (perhaps they get before it dead)

p3n3m:
Check [http://onlineemailvalidation.com/](http://onlineemailvalidation.com/) - it also shows if the domain exists/MX record exists

szepeviktor:
+1 for Travis CI

martenson:
This is an interesting problem. I think it could be sum up to decision in between:

> once disposable always disposable unless proven otherwise

which creates *weaker* and *longer* list of domain

or 

> only disposable when proven disposable

which creates *stronger* and *shorter* list of domains

martenson:
It seems one of our forks implemented this: https://github.com/Friday811/disposable-email-domains/commit/4adeff5d66f91cd14a4a609b8fdf4211a033b008

cc @deguif

deguif:
Interesting.

I will be fine with adding a new file which contains only domains that are existing. 
But I will keep the `disposable_email_blacklist.conf` as it is currently and will not drop anything in it, for the reason @irvin gave in this conversation.

PR #437, opened 2023-12-20
added emails from throwawaymail.com

TravisLittlechilds:
from https://www.throwawaymail.com/en


haumacher:
How did you find out that all these domain names are used by throwawaymail.com?

haumacher:
Hi @TravisLittlechilds, I'm wondering how you found all these domains, since the site throwawamail.com does not allow the user to select a domain from a list and has strong protection against automatic queries. However, when I visited the site, I got an e-mail from domain "mybx.site", which is not in your list...

TravisLittlechilds:
@haumacher it was an annoying process over a few weeks consisting of *many* sessions with incognito windows. I didn't detect any kind of pattern in the domains it handed out, so I'm not surprised I didn't manage to find them all.

haumacher:
@TravisLittlechilds I'm still wondering how to (a) automatically decide whether a PR to this list is valid and (b) how a good heuristics would look like that decides whether a new e-mail domain seen in some user registration form is disposable or not (given this list of "well-known" disposable domains).

There are several services that claim to decide about e-mail domains to be disposable or not. On of them (https://check-mail.org/) seems to take the DNS MX record into account that is associated with the domain. 

This "check" applied to the domains you entered gives the following result:

No MX record:

> em4.catchservers.com
> mx4.catchservers.net

Mail server 164.90.194.37

> jual.me
> seoph.website

All other domains have the mail servers 165.22.201.68 and 137.184.154.224

Maybe the maintainers of this list could give some more insights?

martenson:
to my knowledge a valid MX record is not needed for receiving mail

https://datatracker.ietf.org/doc/html/rfc5321#section-5

we've had some discussions about this topic at https://github.com/disposable-email-domains/disposable-email-domains/issues/84 and https://github.com/disposable-email-domains/disposable-email-domains/issues/58

haumacher:
OK, not having a MX record means the domain/host is its own mail server. So not having a MX record finally provides no clue, whether the domain could be a disposable e-mail domain or not. 

But having a MX record resolving to the same IP address as some other "well-known" disposable domain provides some evidence that this domain is also a disposable one, right?

haumacher:
Or the other way around, if a domain has no MX record, but its resolved IP address points to some mail server used by some other "well-known" disposable domain also provides some evidence that this domain is disposable, too.

In the example above, em4.catchservers.com points to 137.184.154.224, which is the same as one of the mail servers of all other domains. mx4.catchservers.net points to 165.22.201.68, which is the other mail server of all other domains entered in this PR.

haumacher:
@martenson I opened issue #456 with initial thoughts for such heuristics.

nillpoe:
@martenson I think this PR should be closed as **irreproducible** for now.

martenson:
@nillpoe The original PR yes, the discussion here is valuable though, but it should live in the 456 linked above. Thanks @haumacher !

PR #454, opened 2024-02-27
Add closetab.email domains to the blocklist

gauz-kodex:
Domains from closetab.email aka 10minute-email.com


nillpoe:
image
image

@martenson I think this PR should be closed

martenson:

PR #452, opened 2024-02-25
Add even more (all?) temp-mail.org domains

haumacher:
This is a Sisyphean task: Use verifymail.io with `huizk.com` from #445: https://verifymail.io/domain/huizk.com

This reports "Related Domains", which can be associated to the same fake-mail service. This result can be verified by comparing MX records and SMTP host addresses. All domains use one of the following SMTP servers: `143.244.201.82` or `143.244.202.96` with various associated host names:

  "mail.ebuthor.com",
  "mail.gexige.com",
  "mail.huizk.com",
  "mail.lendfash.com",
  "mail.massefm.com",
  "mail.minhlun.com",
  "mail.molyg.com",
  "mail.ricorit.com",
  "mail.scorebog.com",
  "mail.snoodi.com",
  "mail.tvp8.com",
  "mail.unitsade.com",
  "mx.mail-data.net"

PR #455, opened 2024-02-29

Madriix:
All the domains come from https://v erifym ail.io/domain/minhlun.com, from the 'Related domains' section. Whenever someone signs up on my website with a disposable email address, it always appears on https://ve rifym ail.io/, and it's always from temp-mail.org as the email provider. They don't appear anywhere on any lists on GitHub. They should be added..


haumacher:
Did the same in #452 with an additional filtering for dead domains.

Madriix:
They still have one: azduan.com
it is not visible in our modifications

The same user also tried with: lettershield.com (source unknown)

PR #458, opened 2024-03-08
Added ebuthor and secretmail domains to blocklist

rosenbergd:
I'm not certain where these are being generated from, but they're marked as disposable here:

- [SecretMail](https://verifymail.io/domain/secretmail.net)
- [Ebuthor](https://verifymail.io/domain/ebuthor.com)


nillpoe:
@martenson I think this PR should be closed as **irreproducible**.

martenson:

PR #473, opened 2024-04-07
Added new temp email domains and alphabetized the list

adamfisher:
Unfortunately I found this list after I compiled my own so I don't have the original places I grabbed all of these from. I opened up the top two pages for temp email in google search and went through adding all the newest ones I could find. If you close it, so be it. I intend to use a merged list of what you have here and my own. Just trying to give back.


martenson:
@adamfisher I appreciate the thought and effort, thank you. I will not merge the PR at this time, since we are aiming to validate the majority of the entries that go in, and these are not sourced. Glad you found our list useful though.

PR #483, opened 2024-05-30
Add additional alt domains for 1secmail.com

joelweinberger:
Add dpptd.com and rteet.com to the blocklist since they both redirect to https://www.1secmail.com/, which is on the blocklist


martenson:
thanks @joelweinberger !

PR #507, opened 2024-08-22

fra-iesus:
added domain used by 10minutemail.net


nillpoe:
@martenson could you close this unmaintained PR?

martenson:

PR #510, opened 2024-09-02
Add dewuch.com & its catch-all version, plus mailmenot.io

karolyi:
We have plenty of bot registrations about dewuch.com and its subdomains, you can look it up, the domain and its subdomains are used as a disposable email service. For example:

https://www.stopforumspam.com/domain/lap.dewuch.com

Also added mailmenot.io.

### DIFF
--- a/disposable_email_blocklist.conf
+++ b/disposable_email_blocklist.conf
@@ -1,3 +1,4 @@
+.dewuch.com
 0-mail.com
 027168.com
 0815.ru
@@ -776,6 +777,7 @@ dev-null.gq
 dev-null.ml
 developermail.com
 devnullmail.com
+dewuch.com
 deyom.com
 dharmatel.net
 dhm.ro
@@ -2037,6 +2039,7 @@ mailme.gq
 mailme.ir
 mailme.lv
 mailme24.com
+mailmenot.io
 mailmetrash.com
 mailmoat.com
 mailmoth.com


martenson:
@karolyi I see mailmenot as a valid entry but I do not know how to create an address with @dewuch. Please explain.

karolyi:
I don't know either. All I know is that I get registrations from random domains under that one. I also served proof that it is a spammer domain by others' experiences as well.

martenson:
We do not merge with other lists, we prefer to validate the entries. I added the one I was able to verify. Thanks @karolyi !

karolyi:
I'm not asking to 'merge with other lists', I ask to add a domain that is clearly used for abuse, which I made a case for plenty of times. Here's a couple more:

https://www.ipqualityscore.com/domain-reputation/dewuch.com
https://www.stopforumspam.com/domain/oh.dewuch.com
https://www.stopforumspam.com/domain/live.dewuch.com
https://www.urlvoid.com/scan/dewuch.com/
https://cleantalk.org/blacklists/149.71.180.192 (see "Spam attacks, abuse log")

If this doesn't prove the merit of my contribution about this spammer domain, I don't know what will.

I might have to fork this repo and use it in my own py3-validate-email module with the domain added, it seems.

martenson:
@karolyi afaik nothing you are showing here proves the domain is used to offer disposable email service which is what this list is for. 

Please feel free to fork the repo and use it in any way you'd like. That's what I made it for.

PR #516, opened 2024-09-25
Add a pointer to the resolve-email npm package

bcomnes:
I integrated the disposable-email-domains into my email resolution module a while back and it is working well. I would love to add a pointer to it in the README here. I also added descriptions two the two links.

### DIFF
--- a/README.md
+++ b/README.md
@@ -138,7 +138,10 @@ async function isDisposable(email) {
 }
 ```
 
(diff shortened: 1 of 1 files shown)


martenson:
@bcomnes I think a link to one npm package (that already does not do any heavy lifting) is enough for our readme. I am happy you find the project useful though, cheers.

PR #523, opened 2024-10-25
Add hthlm.com to blocklist

toddstoker:
hthlm.com is a https://10minutemail.com/ disposable address.

hthlm disposable


martenson:
thanks @toddstoker !

PR #527, opened 2024-11-08
Add anypng.com

steven-tey:
Spotted on https://temp-mail.org/en

CleanShot 2024-11-08 at 14 50 02


martenson:
thanks @steven-tey

PR #529, opened 2024-11-16
remove letmeinonthis.com

derenrich:
domain is no longer used for this. thanks!

### DIFF
--- a/disposable_email_blocklist.conf
+++ b/disposable_email_blocklist.conf
@@ -1858,7 +1858,6 @@ legalrc.loan
 lellno.gq
 lenovog4.com
 lerbhe.com
-letmeinonthis.com
 letthemeatspam.com
 lez.se
 lgxscreen.com


martenson:
@derenrich could you please give me some basic proof of this claim?

derenrich:
@martenson yeah of course i'd be glad to. if you go to https://www.nslookup.io/domains/letmeinonthis.com/dns-records/mx/ you'll see there is no MX record.

hope that helps.

martenson:
@derenrich per standard valid MX records are not required to receive mail -- A record is used as a backup, see https://github.com/disposable-email-domains/disposable-email-domains/issues/84#issuecomment-307558509

Is this your domain

derenrich:
@martenson no it's not my domain. i was just going through the list and noticed this domain was out of date. 

it looks like letmeinonthis.com doesn't have an A record either. it only has one active DNS entry https://www.nslookup.io/domains/letmeinonthis.com/dns-records/ which doesn't seem related to mail delivery at all.

thanks for looking into this.

martenson:
@derenrich that's fair, thanks for looking into this, I'll merge it now

PR #537, opened 2024-11-24
Added kazvi.com to blocklist

Wild1145:
See https://verifymail.io/domain/kazvi.com for confirmation this is a disposable e-mail.


martenson:
@Wild1145 thanks for your suggestion but we do not merge with other lists like this. We want to verify the domain is actually used by a disposable provider by ourselves.

PR #552, opened 2025-01-19
added two domains fanicle.com and fundapk.com

lookquad:
Found two domains and for reference here are the screenshots- 
1. fundapk.com

fundapak com

2. fanicle.com

fanicle com

Thank You!


martenson:
thanks @lookquad !

PR #553, opened 2025-01-22
Add kimasoft.com

mlissner:
Just submitting this to get the ball rolling. It's also blocked by verifymail.io: 

image

I don't know how to create an email with it though.

(edited b/c I accidentally did wrong url at first)


martenson:
@mlissner thanks for your suggestion but we want to verify the domain is actually used by a disposable provider by ourselves.

PR #558, opened 2025-02-06
adding decep.com

tassoman:
decep.com

https://www.usercheck.com/domain/decep.com


martenson:
thanks for your suggestion but we do not merge with other lists like this. We want to verify the domain is actually used by a disposable provider by ourselves.

tompec:
image

martenson:
@tompec perfect, thanks!

tassoman:
https://tempmail100.com/

PR #560, opened 2025-02-13
added youxiang.dev

benbonn:
creation of email via https://youxiang.dev/mailbox


martenson:
@benbonn thanks for the contribution, however you're adding to the wrong list, plus this entry is already present: https://github.com/disposable-email-domains/disposable-email-domains/blob/c352e96b29b5be433a7b856e8767638d1859b608/disposable_email_blocklist.conf#L3954

PR #566, opened 2025-03-03
egvoo.com is a new one that temp-mail.org is using

umuterturk:
image
egvoo.com is the new one.


martenson:
@umuterturk thanks mate!

PR #567, opened 2025-03-08
New domain from temp-mail.io

farhanmasud:
Found a new domain that is used by temp-mail.io

https://www.ipqualityscore.com/domain-reputation/mkzaso.com

Screenshot from 2025-03-09 02-35-19


martenson:
thanks @farhanmasud !

PR #569, opened 2025-03-19
adding corhash net

tassoman:
## Evidence

immagine

### DIFF
--- a/disposable_email_blocklist.conf
+++ b/disposable_email_blocklist.conf
@@ -724,6 +724,7 @@ coolandwacky.us
 coolimpool.org
 copyhome.win
 coreclip.com
+corhash.net
 cosmorph.com
 courrieltemporaire.com
 coza.ro


martenson:
@tassoman thanks for your suggestion but we want to verify the domain is actually used by a disposable provider by ourselves. Do you know where we can obtain a disposable email with this domain?

tassoman:
Now, 
```
corhash.net mail is handled by 94 route3.mx.cloudflare.net.
corhash.net mail is handled by 24 route2.mx.cloudflare.net.
corhash.net mail is handled by 3 route1.mx.cloudflare.net.
```

But, during the past weeks, I got an account subscription by fake email address. Maybe MR can be suspended / closed.

tompec:
it's from https://smailpro.com/
image

martenson:
thanks @tompec and @tassoman !

PR #578, opened 2025-03-30
Add dizigg.com

smiba:
image


martenson:
thanks for these @smiba !

PR #603, opened 2025-06-03

wesleyschlenker:
Disposable emails with these domains can be created via the Temporary Email Pro mobile app on the Apple app store.

I can't take screenshots of the domains within the app, since I'm not paying for the service, but this is how the emails were reportedly created, and I confirmed that with the free verifymail.io service.


martenson:
Hi @wesleyschlenker, thanks for your suggestion but we do not simply merge with other lists. We want to verify the domain is actually used by a disposable provider ourselves. If we cannot do that we do not add them to the list so screenshots or some other means are necessary.

PR #606, opened 2025-06-04
add punkproof.com and mobilesm.com

plucury:
image
image

### DIFF
--- a/disposable_email_blocklist.conf
+++ b/disposable_email_blocklist.conf
@@ -2475,6 +2475,7 @@ moakt.co
 moakt.com
 moakt.ws
 mobileninja.co.uk
+mobilesm.com
 mobilevpn.top
 moburl.com
 mockmyid.com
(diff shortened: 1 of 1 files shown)


martenson:
@plucury thanks for your suggestion but we do not simply merge with other lists. We want to verify the domain is actually used by a disposable provider by ourselves. Do you know where we can obtain a disposable email with this domain?

plucury:
@martenson You can check them with following link

- https://verifymail.io/domain/mobilesm.com
- https://punkproof.com

martenson:
@plucury that is not checking them, that is just using another list as a reference. Where can I create an email address with these domains?

tompec:
punkproof.com is from https://mail.tm

mobilesm.com is an old domain from temp-mail.org, so there won't be any live evidence for it anymore.

image

martenson:
thanks @tompec and @plucury

PR #609, opened 2025-06-06
Add tempmail.world domains

drygdryg:
Tempmail service https://tempmail.world generates the following domains:
```
business-degree.live
cancer-treatment.xyz
credit-loans.xyz
examstudy.xyz
```
image
image
image
image


martenson:
thanks @drygdryg !

PR #616, opened 2025-06-11
imail.edu.vn domains

philipthomasme:
Temp mails using these domains can be created here: https://imail.edu.vn/


martenson:
thanks @philipthomasme

PR #621, opened 2025-06-16
add hidmail.org

huhuhang:
add hidmail.org

20250616-12-13-39-2Vjk9Uwg

### DIFF
--- a/disposable_email_blocklist.conf
+++ b/disposable_email_blocklist.conf
@@ -1705,6 +1705,7 @@ hidebusiness.xyz
 hidemail.de
 hidemail.pro
 hidemail.us
+hidmail.org
 hidzz.com
 highbros.org
 hihi.lol


martenson:
@huhuhang thanks for your suggestion but we want to verify the domain is actually used by a disposable provider by ourselves. Do you know how we can obtain a disposable email with this domain?

huhuhang:
> @huhuhang thanks for your suggestion but we want to verify the domain is actually used by a disposable provider by ourselves. Do you know how we can obtain a disposable email with this domain?

Hi. You can visit `hidmail.org` and add the corresponding Telegram bot account to get a temporary email address. Recently, our website has been targeted by a large number of users who use these temporary email addresses for malicious credit - card registration tests. They disguise themselves through Telegram bots, and the harm they cause is greater than that of traditional temporary email websites.

martenson:
@huhuhang I'm reluctant to create a telegram account for this. Can you provide more screenshots or description of how it works please?

ghost:
@martenson I tested the Telegram bot and I can confirm that the domain `hidmail.org` is theirs and is valid.

Steps to reproduce:
- Send `/start` as a message to the bot `@hideemailbot`
- Join their group or donate $5.00 USD
- Click on the “Create email” button and the bot will generate the disposable email.

Screenshot:
image
image

martenson:
seems legit, thanks @huhuhang @badappearance 
I assume when you delete the channel/thread the email stops being delivered

ghost:
@martenson Yes, you can generate unlimited emails and you can click the “Disable” button to stop the email from working at any time, similar to SimpleLogin, but by being free and unlimited, it can be heavily abused, one could argue that it is a privacy alias and not temporary but by its nature and what they call themselves I think it would still apply as temporary.

https://verifymail.io/domain/hidmail.org Define it as a privacy alias.

image

PR #627, opened 2025-06-23
feat(domains): Add adguard.com temp mail domain

verticalsync:
This is a domain from adguard's free temp mail service

image
https://adguard.com/en/adguard-temp-mail/overview.html


martenson:
thanks @verticalsync !

PR #631, opened 2025-06-25
feat(domains): add burnerinbox.net domain

ghost:
When entering https://burnerinbox.net/ the only domain that appears is `binbx.net` when generating a new address.

- binbx.net
image


martenson:
@badappearance danke!

PR #633, opened 2025-06-26
feat(lists): add temp-mail.best domain

badappearance:
When entering https://temp-mail.best/ the only domain that appears is `temp-mail.best` when generating a new address.

- temp-mail.best
image


martenson:
thank you @badappearance !

PR #636, opened 2025-06-28
feat(domain): add temp-mail.org domain

nillpoe:
https://temp-mail.org/

### DIFF
--- a/disposable_email_blocklist.conf
+++ b/disposable_email_blocklist.conf
@@ -1260,6 +1260,7 @@ exclussi.com
 exdonuts.com
 exelica.com
 existiert.net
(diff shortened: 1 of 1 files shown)


ghost:
@nillpoe Make sure to move your domain between `existiert.net` and `exitstageleft.net` as right now it is not correctly sorted.

@martenson Here is the proof of the domain in case you need it:
image

nillpoe:
> @nillpoe Make sure to move your domain between `existiert.net` and `exitstageleft.net` as right now it is not correctly sorted.

done :)

martenson:
thank you both!

PR #660, opened 2025-06-29
Add surfshark.com domain

nillpoe:
image
https://surfshark.com/blog/email-address-generator

### DIFF
--- a/disposable_email_blocklist.conf
+++ b/disposable_email_blocklist.conf
@@ -3248,6 +3248,7 @@ rteet.com
 rtrtr.com
 rtskiya.xyz
 rudymail.ml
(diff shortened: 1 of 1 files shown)


martenson:
I need to check this case closer, surfshark is a legit service, not sure what they're doing here yet.

ghost:
@martenson I can confirm that it is not a disposable email. To use Surfshark Alternative ID, you have to pay for the Surfshark Starter plan or higher, This service is almost identical to SimpleLogin, it's for creating aliases of your original email, it's not a disposable email.

image

martenson:
Thanks @badappearance , good to close @nillpoe ?

PR #667, opened 2025-06-30

BeyondVertical:
Added tca7.com

I know I would have to add where one can generate this, but I do not know.
However, look here - it is definitely recognized as a displosable domain (on multiple of those sites): https://www.usercheck.com/domain/tca7.com

### DIFF
--- a/disposable_email_blocklist.conf
+++ b/disposable_email_blocklist.conf
@@ -3666,6 +3666,7 @@ taukah.com
 taxibmt.com
 tb-on-line.net
 tbgroupconsultants.com
+tca7.com
 tcwlm.com
 tcwlx.com
 tdtda.com


martenson:
@BeyondVertical hi, just as you mentioned we do not merge with other lists blindly, if we cannot validate we won't merge. How did you come by this domain?

BeyondVertical:
User registration in a forum I am administrating. The user is trolling around, so I checked out the domain.

ghost:
After checking, it looks like the following MX records are linked to a disposable email provider:

```
mail.wabblywabble.com
mail.wallywatts.com
```

I wasn't able to determine their exact origin, but it seems to be `temp-mail.org`, as they share the same hosting and domain registrar. However, I'm not entirely sure because these domains are quite different from others they operate. but they are definitely linked to a disposable email generator.

PR #668, opened 2025-06-30
chore: add a new disposable domain from temp-mail.org

wilsonadenuga:
You can generate the domain by going to [temp-mail](https://temp-mail.org/en/)

Screenshot 2025-06-30 at 15 50 38


martenson:
Thanks @oluwatunmiisheii !

PR #669, opened 2025-06-30
Add temp-mail.org domain

nillpoe:
스크린샷 2025-06-30 235423
스크린샷 2025-06-30 235515
스크린샷 2025-06-30 235633
스크린샷 2025-06-30 235718
It seems to determine the domain of the returned email based on each user's 'token' cookie, IP, etc.


nillpoe:
branch collision

PR #687, opened 2025-07-10
feat(domains): add mail.tm domain

verticalsync:
New domain from mail.tm `mechanicspedia.com`

image


martenson:
thanks @verticalsync

PR #691, opened 2025-07-17
Add sites associated with quix.email

bchen-stytch:
We've seen spam/abuse behavior and mass automated account creation attempts from these email domains at Stytch. 

You can purchase disposable emails from these domains via https://quix.email ([FAQ](https://quix.email/faq)). I've filtered out those from commonly-used free domains like Gmail and Hotmail. 

image

image


martenson:
thank you @bchen-stytch !

PR #708, opened 2025-08-18
Add three domains

runbing:
Add three domains as follows, which can be produced on https://temporam.com/ems.

* temporam.com
* mona.edu.kg
* mona.edu.pl

s1

s2

s3


martenson:
thanks @runbing !

PR #743, opened 2025-09-22
Add newaddr.com

MikeW1901:
newaddr.com - abuse link on homepage goes to https://magical.kuku.lu/page.contact.php?type=MailNow which is the support form for InstAddr, a disposable email provider.

image

image

http://133.130.102.20/ (MX record IP) redirects to https://m.kuku.lu/en.php also.


martenson:
thanks @MikeW1901

PR #746, opened 2025-09-23
Add safetoca.com and salexup.com

Ohnoimded:

### DIFF
--- a/disposable_email_blocklist.conf
+++ b/disposable_email_blocklist.conf
@@ -3490,11 +3490,13 @@ saeoil.com
 safaat.cf
 safermail.info
 safersignup.de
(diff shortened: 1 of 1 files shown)


nillpoe:

Ohnoimded:
https://www.usercheck.com/domain/salexup.com

https://www.usercheck.com/domain/safetoca.com

image

image

nillpoe:
That is other list too.

Ohnoimded:
Sorry, I don't it would be possible without requesting new emails from temp-mail and cross checking that until I land on these mails.

nillpoe:
@martenson please close this PR since it's irreproducible.

PR #747, opened 2025-09-23
Add exiond.com

alrocha:
Checked on https://verifymail.io/domain/exiond.com
Screenshot 2025-09-23 at 23 27 49


nillpoe:
thanks for your suggestion but we do not simply merge from the other lists. We want to verify the domain is actually used by a disposable provider by ourselves. We need a screenshot or similar.

nillpoe:
otherwise, close this PR please.

alrocha:
Is this enough? https://www.ipqualityscore.com/domain-reputation/exiond.com

nillpoe:
That is other list too.

alrocha:
Understood, thanks anyway!

PR #755, opened 2025-10-03
Add latest priyo.email domains

bchen-stytch:
Latest domains list from https://priyo.email - this is the set of domains available when you select the "Change" button and look at the domains options.

image

image-6


martenson:
awesome, thanks @bchen-stytch !

Issue #665, opened 2025-06-30
adapt code examples to SLDs


martenson:
We did not update code examples properly after adapting to SLDs. They still work but some have false negatives.

I think we should match from the right and if the full record in the blocklist matches the query then the address is disposable no matter what precedes it.

xref https://github.com/disposable-email-domains/disposable-email-domains/pull/649#discussion_r2174255968

PR #760, opened 2025-10-13
Update Python Example to Handle Third Level Domains and Higher

### LINKED ISSUE #759: Disposable Email Subdomains
Hi I was wondering how disposable domains + subdomain should be handled.

I've found the following domains + subdomain in use on [tempmail.lol](https://tempmail.lol)
```
*.chessgameland.com
*.chessgamingworld.com
*.vertexium.net
*.191mariobet.com
```
Image
Image
Image
Image

Considering the code from the README (`email.partition('@')[2] in blocklist_content`), we would need to list each subdomain (which doesn't seem feasible given the amount).

I know disposable-email-domains doesn't merge lists but I think [this report from usercheck](https://www.usercheck.com/provider/tempmail.lol) can help provide some proof. There are a lot of these *.[domain].[top-level] being added within the last couple of hours.

Happy to open a PR with any changes you may suggest!

andrewbaldwin44:
### Proposal
- Add a new blocklist to check against wildcard domains
- Update the verify.py script to filter against the public suffix list (e.g. *.co.uk)
- Update the README.md to provide a basic python example of how the new blocklist can be used. Happy to update a couple of the other language examples that I am comfortable with if this is accepted

Fixes #759

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,7 +58,6 @@ for i in range(len(domain_parts) - 1):
     if ".".join(domain_parts[i:]) in blocklist_content:
         message = "Please enter your permanent email address."
         return (False, message)
-        
 return True
 ```
 


anwarjaved:
I don't think wildcard is needed because when you check endWith its automatically covers it

andrewbaldwin44:
Searching from the right sounds nice but I think it could come up with far too many false positives. E.g. I wouldn't be able to have a domain like "great-boats.us" if "boats.us" were to be in the blocklist

martenson:
@andrewbaldwin44 As I see it every domain in the list is considered a wildcard domain. I.e. once you control the second level, you control the third level and deeper. 

Matching from the right irrespectively of the domain name is a bad idea.

andrewbaldwin44:
@martenson You mentioned wanting to match from the right correct? #665

If we want to treat every domain in the list as a wildcard I would suggest checking for either
A. An exact match in blocklist content
B. Matching from the right including a dot e.g. `endswith(f".{blocklist_entry}")`

Some of the entries in the blocklist content are very short, for example "99.com", which I imagine could end up with a lot of false positives

martenson:
@andrewbaldwin44 it's been a while but I think that comment is meant "to match the _full_ query", not any subquery -- that feels very broken. If I meant it that way, it was a brainfart.

andrewbaldwin44:
Ah I think I understand you now. 
I've updated the python example now to match each part of the email domain:
```
-> example.api.mail.com in blocklist_entry
-> api.mail.com in blocklist_entry
-> mail.com in blocklist_entry
-> no matches!
```
This will keep our O(1) optimization

If this looks good to you I can update some of the other examples and I'll add my domains in another PR. Thanks!

README.md
```diff
@@ -52,11 +52,14 @@ TOC: [Python](#python), [PHP](#php), [Go](#go), [Ruby on Rails](#ruby-on-rails),
...
-    return (False, message)
-else:
-    return True
+
+domain_parts = email.partition('@')[2].split(".")
+for i in range(len(domain_parts) - 1):
+    if ".".join(domain_parts[i:]) in blocklist_content:
+        message = "Please enter your permanent email address."
+        return (False, message)
+
```
martenson:
```suggestion
```

martenson:
this feels like an improvement, thank you @andrewbaldwin44

PR #762, opened 2025-10-16
Add apocaw.com to disposable email blocklist

abdoul-rb:
This email address can be generated from the website temp-mail.org/

apocaw-temp-mail


martenson:
thanks @abdoul-rb

PR #766, opened 2025-10-21
fix: remove gynzy domains

joelluijmes:
Gynzy is an educational platform based in the Netherlands. We’re not sure how we ended up on this list, but please consider removing us 🙏

More info: https://www.gynzy.com/en/ happy to provide additional details if needed 😇


martenson:
@joelluijmes sorry for the inconvenience

PR #773, opened 2025-10-31
Feat: Fetching directly from Disposable domains

Jotenks66:
Added a Github workflow to automatically (daily) fetch disposable email domains. 
Added Yopmail but created it in a way that new domains can be added to the same file.

Also added hh7f.com to the list from https://temp-mail.org/en/

image


nillpoe:
nice

martenson (approved):
awesome, let's give it a shot!

martenson:
@Jotenks66 results of the first run are here: https://github.com/disposable-email-domains/disposable-email-domains/pull/778

It seems to work. There are a couple of third level domains but they seem to be on the PSL so the entries are afaik valid.

Thanks so much!

PR #774, opened 2025-11-01
Add easymails.cc disposable email domain

Salzling:
You can generate a disposable email by going to https://easymails.cc/.

Screenshot below:
A08Q4sh1yi


martenson:
thanks @Salzling !

PR #803, opened 2025-11-17
adding limtu.com

malwador:
Add [limtu.com](http://limtu.com/)

https://www.usercheck.com/domain/limtu.com


nillpoe:
Hey @malwador, thanks for your suggestion but we do not simply merge with other lists.
We want to verify the domain is actually used by a disposable provider ourselves.
If we cannot do that, we do not add them to the list, so screenshots or some other means are necessary.

nillpoe:
Hey @malwador, if there's no one, close the PR please!

PR #806, opened 2025-11-20
Update to include emails from temp-mail.org

mmcc:
Found a signup come through and traced the email back to [temp-mail.org](https://temp-mail.org). Iterated a few different emails in an incognito window and found a few more that weren't in this list. I didn't grab a screenshot before closing okcdeals.org and got rate limited when going back for screenshots of the others, but unfortunately there isn't a way to see a list of all available domains and you just have to keep loading the page in an incognito window.

image
image
image
image

### DIFF
--- a/disposable_email_blocklist.conf
+++ b/disposable_email_blocklist.conf
@@ -597,6 +597,7 @@ bio-muesli.info
 bio-muesli.net
 biojuris.com
 bione.co
+bipochub.com
 bitmens.com
 bitwhites.top
 bitymails.us


mmcc:
Added bipochub.com, and thanks for the screenshot showing okcdeals.com (that was already in the PR, I just wasn't able to get a screenshot).

disposable_email_blocklist.conf
```diff
@@ -3188,6 +3193,7 @@ ohm.edu.pl
 oida.icu
 oing.cf
 okclprojects.com
+okcdeals.com
```
nillpoe: 'okcdeals.com' should come before 'okclprojects.com'

nillpoe:
@mmcc please sort the lines correctly!

wdm-vii:
@mmcc I found that `datehype.com` is also maintained by temp-mail.org
https://www.usercheck.com/domain/datehype.com
There is also a list of the other domains maintained by them [here](https://www.usercheck.com/provider/temp-mail.org)

PR #810, opened 2025-12-04
Remove altmails.com

nillpoe:
image

not a disposable temporary email provider


martenson:
thanks for the update @nillpoe

PR #813, opened 2025-12-09
Revert "Remove altmails.com"

martenson:
Actually, they're still doing it. @nillpoe 

Screenshot 2025-12-09 at 12 31 37 PM
Reverts disposable-email-domains/disposable-email-domains#810

### DIFF
--- a/disposable_email_blocklist.conf
+++ b/disposable_email_blocklist.conf
@@ -363,6 +363,7 @@ alpinewe.us
 alreval.com
 altairwe.us
 altitudewe.us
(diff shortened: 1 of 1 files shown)


nillpoe:
Oops, I thought they're a legitimate email relay provider.

nillpoe:
is altmails.com just a cloak?

martenson:
@nillpoe dunno what cloak is, and they can provide legit service too, but once they provide also a disposable email service they get on the list I think

PR #814, opened 2025-12-09
Correct Ruby example

andrewbaldwin44:
`disposable_email?` should return `true` when disposable, otherwise `false`


martenson:
thanks @andrewbaldwin44 @oshanz!

PR #836, opened 2025-12-27
mail.tm domain is now supporting mail.gw

SaiAkashNeela:
image

mail.tm disposable service now supports mail.gw (https://mail.gw/en/) domain. Sometimes its generating emails like user@mail.gw, attached screenshot and URL for reference.


martenson:
thanks @SaiAkashNeela !

PR #837, opened 2025-12-28
New disposable domains added

LautBox:
New disposable email domains added

Domains added: 49 domains from 5x2.de service
(see full list in the file changes)

Main domains include:
- 5x2.de, 5x2.me, 5x2.email
- Various German fun domains (*.sexy, *.bar, *.bayern, etc.)

Proof: These domains are offered as disposable email service at https://5x2.de
Screenshot verification attached showing the available domain list.

Note: Unable to execute maintain.sh locally. 
Please run the script to format (lowercase, sort, remove duplicates).

30324ed7-fbdc-4544-9bfc-edb9cf3c63b5


nillpoe:
please sort the lines correctly

nillpoe:
also 5x2.de seems legitimate to me (email alias provider).
could you confirm that if we need an existing email address to use their alias ?

LautBox:
> also 5x2.de seems legitimate to me (email alias provider). could you confirm that if we need an existing email address to use their alias ?

Thank you for reviewing!

You're right that 5x2.de requires an existing email address for registration. 
However, I believe these domains should still be blocked for the following reasons:

1. Domain names clearly indicate spam/disposable usage:
   - keinspam.email (German: "no spam")
   - mag-spam.net (German: "likes spam")
   - nurspam.email (German: "only spam")
   - will-keinen-spam.de (German: "don't want spam")
   - ohne-drogen-gehts.net (German: "works without drugs")

2. Service appears abandoned: The send function is integrated but doesn't work, and the service seems unmaintained for years.

3. Primary use case: These email addresses are designed and used for spam/junk registrations, not as legitimate email aliases. The domain names themselves advertise this purpose.

However, I understand if you disagree. These domains sit in a gray area between email alias and disposable email. I'm happy to close the PR or reduce it to only the most obvious spam-related domains if you think that fits better.

What's your opinion?

nillpoe:
> Service appears abandoned: The send function is integrated but doesn't work, and the service seems unmaintained for years.

But you can still receive an email with that alias address ?

nillpoe:
https://pastes.dev/V48ePMWsaJ
this is a fully sorted one btw

LautBox:
> > Service appears abandoned: The send function is integrated but doesn't work, and the service seems unmaintained for years.
> 
> But you can still receive an email with that alias address ?

Yes, emails can still be received with these alias addresses. They forward to the user's registered email address.

nillpoe:
thanks for confirming that.

but i think email alias provider without sending function is very general

LautBox:
> but i think email alias provider without sending function is very general

You're absolutely right. I understand now - just because the sending function is missing doesn't make it a disposable email service. Since receiving still  works, these are functioning email aliases.

Thank you for taking the time to review this and for the sorted list! 
I've learned a lot from this discussion. I will close this PR.

PR #842, opened 2026-01-05
Use mktemp so we don't write to disk

yigib:
This script unnecessarily writes to disk. 
Could've been a one-liner but that would sacrifice readability.
mktemp is a good compromise. Keeps script short and is pretty portable (available on GNU Core Utils, BusyBox, OpenBSD, FreeBSD)


martenson:
@yigib I suspect you don't know where tmp is mounted, so this could still write to disk. It is never worse though. Thanks!

PR #844, opened 2026-01-06
Add mypup.nl to allow list because it't not a (disposable) email provider at all

DennisMyPup:
MyPup is a parcel delivery company and does not provide disposable email addresses.

Unfortunately we're being harassed at the moment and someone put our domain name mypup.nl up for disposable e-mail domain. We're still investigating where it comes from, but it would be nice if it could be on this allow list.

### DIFF
--- a/allowlist.conf
+++ b/allowlist.conf
@@ -125,6 +125,7 @@ mm.st
 mozmail.com
 myfastmail.com
 mymacmail.com
+mypup.nl
 naver.com
 neverbox.com
 nigge.rs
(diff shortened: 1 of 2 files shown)


4-fa:
@nillpoe Related issue: https://github.com/disposable-email-domains/disposable-email-domains/issues/845

martenson:
all poisoned entries removed in https://github.com/disposable-email-domains/disposable-email-domains/pull/848

Thank you for the help @4-fa, @nillpoe  and sorry for the inconvenience @DennisMyPup

PR #852, opened 2026-01-13
Add Additional Verification Records for tmailor.com Domain

Manas1820:
### Temporary Email Domain Verification

- The email domain was generated using a temporary email service:  
  https://tmailor.com/temp-mail

- To verify the domain reputation and classification, you can independently review it here:  
  https://www.usercheck.com/domain/hetzez.com
  
### Edit

After adding this, it seemed appropriate to include all related verification details for completeness.

PR #853, opened 2026-01-13
Add additional domains to the disposable email blocklist

Manas1820:
Here are the valid proofs for reference.
Please see #852  for additional context.

I have valid proof for some additional entries and would like to include those.

aliban org
ket-qua org
pongpong org
tikmail org
4save net
contaco org
kontoko org
libinit com


Manas1820:
@martenson this is a brach out from #852 , this contain valid ss as proof for some specific domains

martenson:
thank you @Manas1820 !

PR #857, opened 2026-01-15
Add awesomesaucemail.org to blocklist

gopi-ar:

https://tempmail.lol/en/

This website rotates between the following domains, 2 of which are already on the list:
chessgameland.com
awesomesaucemail.org
chessgamingworld.com

They also generate emails with random subdomains on these 3 parent domains. Should we add a note in the README about matching only parent domains if that would help others avoid skipping these domains?

### DIFF
--- a/disposable_email_blocklist.conf
+++ b/disposable_email_blocklist.conf
@@ -507,6 +507,7 @@ awa.pics
 awatum.de
 awdrt.org
 awesome47.com
+awesomesaucemail.org
 awgarstone.com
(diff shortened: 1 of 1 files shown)


martenson:
@gopi-ar thanks! I verified the new entry, next time please include screenshot

> Should we add a note in the README about matching only parent domains if that would help others avoid skipping these domains?

There is a note in the readme about this. We only block second level domains unless they use public suffix.

PR #858, opened 2026-01-15
Update PHP example to support subdomain matching

gopi-ar:
The PHP example doesn't handle subdomains - it only checks exact domain matches. This update brings it in line with the other language examples.


martenson:
thank you @gopi-ar !

PR #862, opened 2026-01-20
Add relay.mozmail.com to whitelist

Digital-Toxin:
relay.mozmail.com is used by Firefox Relay for authenticated email alias forwarding.

Since mozmail.com is already whitelisted and Firefox Relay does not offer disposable or publicly accessible inboxes, this subdomain should be treated the same.


martenson:
we only capture second level domains

PR #865, opened 2026-01-27

hazaveh:
adds to blocklist coffeeity.com


martenson:

hazaveh:
I got an email under this domain with one of these temp providers. unfortunatly I do not remember which one. but Yes I personally got one and realized its not listed in the DB.

martenson:
@hazaveh I am sorry but that is not a proof of disposability

PR #866, opened 2026-01-28
Add Elixir example 🧙 ⚗️

oshanz:
Added an Elixir example for checking disposable emails, along with an Elixir package.


martenson:
lgtm and the llm I asked for elixir syntax check :D, thanks @oshanz

PR #867, opened 2026-01-28
Add ixunbo.com, eldobhato-level.hu to disposable email blocklist

icetee:
Adds to blocklist ixunbo.com, eldobhato-level.hu

image

image

image


martenson:
thanks @icetee !

PR #876, opened 2026-02-04
add more deepmails.org domains

miketheman:
Some of their existing entries are here, this expands their list to their current avaialble entries (excluding gmail.com)

Screenshot 2026-02-04 at 12 23 42


martenson:
thank you @miketheman !

Issue #872, opened 2026-02-02
add dnsclick.com


chandhravadanan:
https://temp-mail.org/en/

<img width="1512" height="552" alt="Image" src="https://github.com/user-attachments/assets/14687fd0-bfd8-4c9b-a14d-af97158fb122" />

chandhravadanan:
https://github.com/disposable-email-domains/disposable-email-domains/issues/588

KroZen-Dev:
<img width="1605" height="684" alt="Image" src="https://github.com/user-attachments/assets/c68ed123-a05a-43e6-bf7f-e0d0fff2fa9c" />

KroZen-Dev:
<img width="1657" height="496" alt="Image" src="https://github.com/user-attachments/assets/6ab365ce-bd2d-4ef7-ae55-d5998ebfb342" />

KroZen-Dev:
opemails.com
juhxs.com
emlhub.com
creem.me
furvionx.com
naprb.com
atinjo.com
gopicta.com
jparksky.com
eubonus.com
mriscanner.live
ixospace.com
akixpres.com
koletter.com
webxio.pro
cucadas.com
gavrom.com
dubokutv.com
hudisk.com
smail.qzz.io
kanonmail.com
gyknife.com
any.pink

https://www.usercheck.com/domain/opemails.com
https://www.usercheck.com/domain/juhxs.com
https://www.usercheck.com/domain/jparksky.com
https://www.usercheck.com/domain/akixpres.com
https://www.usercheck.com/domain/ixospace.com

martenson:
@chandhravadanan would you care to create a PR for the `dnsclick com`?

@KroZen-Dev thanks for your suggestion but we do not simply merge with other lists. We want to verify the domain is actually used by a disposable provider by ourselves. Do you know where we can obtain a disposable email with this domain?

KroZen-Dev:
> [@chandhravadanan](https://github.com/chandhravadanan) would you care to create a PR for the `dnsclick com`?
> 
> [@KroZen-Dev](https://github.com/KroZen-Dev) thanks for your suggestion but we do not simply merge with other lists. We want to verify the domain is actually used by a disposable provider by ourselves. Do you know where we can obtain a disposable email with this domain?

This group of people is attacking hosting providers. They do this without websites or anything. They rent certificates and create their own closed (private) services for themselves; unfortunately, they are not available online :(

martenson:
[...]

PR #878, opened 2026-02-06
fix(872): add dnsclick.com

chandhravadanan:
https://github.com/disposable-email-domains/disposable-email-domains/issues/872


martenson:
thanks @chandhravadanan

PR #881, opened 2026-02-13
Add 70 disposable/temporary email domains identified from production abuse patterns

AniketPadmakar:
These domains were identified through analysis of signup patterns on a production SaaS platform over the past 2 weeks. All domains showed bulk registration patterns (8-50 signups each) with nonsensical domain names, characteristic of disposable email services.

Includes:
- Bulk abuse domains (.com): cimario, codgal, aixind, azeriom, etc.
- .cfd TLD domains: deltabox, lyoris, pidge, poststream, relayix, vexisync, xitro
- Various TLD domains: .icu, .world, .live, .shop, .online, .store, .click, .fyi, .vip
- Fake email service domains: gmail10p, mailpwr, oemails, suiemail, mimimail.me, etc.

### DIFF
--- a/disposable_email_blocklist.conf
+++ b/disposable_email_blocklist.conf
@@ -54,6 +54,7 @@
 10minutmail.pl
 10x9.com
 11163.com
+1200b.com
 123-m.com
 123clone.com
 123mails.org
@@ -117,6 +118,7 @@
 2chmail.net
 2ether.net
 2fdgdfgdfgdf.tk
+2insp.com
 2odem.com
 2prong.com
 2wc.info
@@ -319,6 +321,7 @@ air2token.com
 airmailbox.website
 airsworld.net
 aiworldx.com
+aixind.com
 aixnv.com
 ajaxapp.net
 akapost.com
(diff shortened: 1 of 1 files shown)


craigRSA:
could you add webxio.pro to this PR too please?
https://verifymail.io/domain/webxio.pro

martenson:
@AniketPadmakar Thanks for your suggestion but we do not simply merge with other lists. We want to verify the domain is actually used by a disposable provider by ourselves. Do you know where we can obtain a disposable email with these domains? You can also provide a screenshot of the disposable domain service page as a basic proof.

AniketPadmakar:
> @AniketPadmakar Thanks for your suggestion but we do not simply merge with other lists. We want to verify the domain is actually used by a disposable provider by ourselves. Do you know where we can obtain a disposable email with these domains? You can also provide a screenshot of the disposable domain service page as a basic proof.

Hi @martenson, my evidence is behavioral — these domains showed bulk signups (8–50 each in 2 weeks) with bot-like usernames on a production platform. Most don't have a consumer-facing UI; they operate via SMTP relay which is why they're not in existing lists.

martenson:
@AniketPadmakar I do believe you and your analysis, however to get on this list the service has to be user-facing. What you are describing could be a private targeted scammer effort that we have no means to verify now, or in the future. 

Afaik if they're using open mail relay most mail service will not trust them anyways.

PR #883, opened 2026-02-24
add more temp-mail.org domains

Wiola09:
* add more temp-mail.org domains detected after 2026-02-21

See e.g. https://verifymail.io/domain/tempmail.co

### DIFF
--- a/disposable_email_blocklist.conf
+++ b/disposable_email_blocklist.conf
@@ -277,6 +277,7 @@ adventwe.us
 advisorwe.us
 advitise.com
 advocatewe.us
+advarm.com
 adwaterandstir.com
 aegde.com
 aegia.net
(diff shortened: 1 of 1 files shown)


martenson:
@Wiola09 thanks for your suggestion but we do not simply merge with other lists. We want to verify the domain is actually used by a disposable provider by ourselves. Do you know where we can obtain a disposable email with this domain?

Wiola09:
Thanks for the clarification. You can get disposable addresses for these domains at temp-mail.org — all four were created there. They are detected when users tried to register on production system using temp-mail.org addresses, so they are in active use as disposable emails.

martenson:
@Wiola09 you get a random domain at `temp mail org`, please provide screenshots with the specific domains you want to add. This is the basic means of verification we use.

Wiola09:
Understood. I also tried to get those domains on temp-mail.org but couldn’t reproduce them; they appeared in the registration attempts we logged on production. Thanks for the feedback — I’ll close the PR.

PR #886, opened 2026-02-26
Add free-temp-mail.eu.org domains

SamHimmy:
Domains that they use
free-temp-mail.eu.org
sam1.eu.org
temporaryemail.dpdns.org
Screenshot_20260226_075106_Chrome (1)


martenson:
thank you @SamHimmy

PR #892, opened 2026-02-28
chore: prevent maintenance running on forks

miketheman:


martenson:
I feel like forks should manage their own CI but this is also fine. Thanks @miketheman

PR #894, opened 2026-03-03
Add two temp-mail.org disposable domains

Wiola09:
This PR adds two disposable email domains originating from temp-mail.org:

- ostahie.com,
- pazuric.com

These domains were observed in registration attempts on a production system using temp-mail.org addresses. As requested, screenshots demonstrating these domains in temp-mail.org are provided below for verification.

ostahie com
pazuric com


martenson:
thanks @Wiola09 !

PR #907, opened 2026-03-17
Add dollicons.com to disposable email blocklist

alejo2246:
image

### DIFF
--- a/disposable_email_blocklist.conf
+++ b/disposable_email_blocklist.conf
@@ -1218,6 +1218,7 @@ dodsi.com
 dogclothing.org
 doiea.com
 doj.one
+dollicons.com
 dolofan.com
 dolphinnet.net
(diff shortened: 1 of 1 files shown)


martenson:
@alejo2246 Thanks for your suggestion but we do not simply merge with other lists. We want to verify the domain is actually used by a disposable provider by ourselves. Do you know where we can obtain a disposable email with this domain? If so please provide a screenshot of the disposable domain being used there.

alejo2246:
Hi @martenson,

I’m not sure where someone could obtain a disposable email address using this domain, but my app appears to be vulnerable to registrations from disposable email services associated with it.

martenson:
We require a basic proof, presence on other lists is not enough.

alejo2246:
image
It's not in other lists, is on our database.

PR #908, opened 2026-03-18

lana-iakovlieva:
Caught these being used for fake registrations on my site


martenson:
@lana-iakovlieva Please run the maintenance script in this repo to format the new entry properly, we also need some form of proof these are from disposable services.

PR #916, opened 2026-03-21
Add fakeout as alternative NPM package

Manas1820:
## Summary

Hey! I noticed the existing NPM package listed here (`disposable-email-domains-js`) can sometimes take weeks to catch up with changes made to this repo. Since this blocklist gets updated frequently via git, there's a gap between when a domain is added here and when it's available to Node.js users pulling from npm.

[fakeout](https://www.npmjs.com/package/fakeout) solves this by syncing directly with this repo daily via CI/CD and auto-publishing patch releases — so Node.js consumers stay up to date without manual intervention. It also:

- Has zero dependencies (just a `Set` lookup)
- Ships with full TypeScript support
- Provides convenient helpers: `isDisposableEmail()`, `isDisposableDomain()`, `getDisposableDomains()`

Given how actively this repo is maintained, having an npm package that actually keeps pace with it feels like a good addition. Not replacing the existing alternative — just offering another option that prioritizes staying in sync.

### DIFF
--- a/README.md
+++ b/README.md
@@ -163,9 +163,7 @@ function isPermanentEmail(email) {
 }
 ```
 
-Alternatively check out NPM package https://github.com/mziyut/disposable-email-domains-js.
-
-Also check out [fakeout](https://www.npmjs.com/package/fakeout) — a zero-dependency package with an auto-updating blocklist that syncs daily via CI/CD, publishing new patch versions automatically. It provides `isDisposableEmail()`, `isDisposableDomain()`, and `getDisposableDomains()` functions with full TypeScript support.
+Alternatively check out NPM packages https://github.com/mziyut/disposable-email-domains-js (updated weekly) and https://www.npmjs.com/package/fakeout (updated daily).
 
 ### C#
 ```C#


Manas1820:
@martenson Totally understand if this doesn't feel like the right fit  just sharing it since I've been using fakeout with a  Dependabot setup and it's been working well for keeping things in sync automatically

martenson:
@Manas1820 afaik for the node package you can run `npm run generate:disposable_email_blocklist_json` to update -- is that not correct?

Manas1820:
> @Manas1820 afaik for the node package you can run `npm run generate:disposable_email_blocklist_json` to update -- is that not correct?

@martenson yeah, that’s correct I did see that.

The expected flow makes sense, but in practice the repo was running about a week behind upstream updates.

I didn’t want to make changes to the existing setup in this repo, so I built a separate approach to keep things in sync automatically.

martenson:
@Manas1820 mhm, you're right, so your solution basically adds a daily cron on top of it? Wouldn't it be less fragmented adding it as a PR to https://github.com/mziyut/disposable-email-domains-js then?

Manas1820:
> @Manas1820 mhm, you're right, so your solution basically adds a daily cron on top of it? Wouldn't it be less fragmented adding it as a PR to https://github.com/mziyut/disposable-email-domains-js then?

@martenson Yeah, that’s fair I can definitely try raising a PR for it.

My only concern is around how active the maintenance is, since the lag we’re seeing suggests updates aren’t happening very frequently. Although I do see Dependabot updates coming in so mostly it looks like its just updates not being merged daily level.

That’s also why I initially kept this as a separate setup.

martenson:
If I read it correctly it was at least 3 updates in March, so roughly 1 every week. Honestly that feels above and beyond for this kind of a list.

https://github.com/mziyut/disposable-email-domains-js/commits/main/

Manas1820:
> If I read it correctly it was at least 3 updates in March, so roughly 1 every week. Honestly that feels above and beyond for this kind of a list.
> 
> https://github.com/mziyut/disposable-email-domains-js/commits/main/

Yep, it’s on a weekly update cycle that’s why I mentioned this. The idea was just to automate things a bit more.

We noticed some disposable emails were already in the upstream list but took time to reflect here. By then, those emails had already been used to create accounts, which meant extra work on banning/flagging and cleanup.

So I put together a quick setup to reduce that gap and limit those cases. Totally fine either way just wanted to share in case someone else has a similar requirement.

Also, feel free to close this if it’s not adding much value here.

martenson:
@Manas1820 I've added the link to your package, thanks for contributing!

Manas1820:
Thanks @martenson  for adding the link, really appreciate it! 🙌
Hopes this helps the community !

PR #936, opened 2026-04-12
remove allowlist and the attached logic

martenson:
this feels no longer relevant, we do a decent job of verifying all new entries in the blocklist and the allowlist seems like an unmaintained dead end list


jgonggrijp:
Thanks for critically verifying additions. For my understanding, how do you do it? I see a forwarding service that I rely on listed in the former allowlist, so the fact that the allowlist is no longer there as a safeguard does make me a little bit nervous. I already get a lot of trouble from other blocklists that mistake it for a burner email service, so I dread the possibility that someone proposes to add it here as well and it ends up being added to the blocklist.

martenson:
@jgonggrijp generally we directly obtain the disposable email address with the domain as a proof. The allowlist was a feature that helped us in the past, but given nothing ever leaves the allowlist it was becoming more of an exploit surface imho.

PR #940, opened 2026-04-12
inserted emailax.pro

DevAbdoTolba:
There is a new domain was added and being used in alot of sites! found it and did test it many times and it did pass through. found this library that many uses it and trusts it. but looks like it had a small hole that the spammers got over :)

### DIFF
--- a/disposable_email_blocklist.conf
+++ b/disposable_email_blocklist.conf
@@ -1406,6 +1406,7 @@ emailage.tk
 emailaoa.pro
 emailate.com
 emailawb.pro
+emailax.pro
 emailbbox.pro
 emailbin.net
 emailboxer.one


martenson:
@DevAbdoTolba Thanks for your suggestion but we want to verify the domain is actually used by a disposable provider by ourselves. Do you know where we can obtain a disposable email with this domain? If so please provide a screenshot of the disposable domain being used there.

DevAbdoTolba:
@martenson 
I do know a site that shares it (I can't share the url here or I would get banned by GH)
I am trying to access it but looks like it's servers are down at the moment!
I will provide a screen shot and a domain just as soon as it is up again 👍

DevAbdoTolba:
Here is an email with the same domain, you might send it a message and it will be delivered

57dbccbb0b@emailax.pro

martenson:
thank you @DevAbdoTolba

PR #942, opened 2026-04-14
Add lealking.com to disposable email blocklist

aviscido:
lealking.com emails are created using a disposable email service known as MailTicking, which can be either reached at mailticking.com or by install the MailTicking app from the either the Apple App Store or the Google Play Store. Difficult to get a screenshot as this service uses a large number of domains.


ishaansehgal99:
+1

martenson:
@aviscido all domains I see at `lealking` are already in this list, if you see a different one please provide a screenshot as a basic proof, thank you

PR #945, opened 2026-04-18
Add new disposable email domains to blocklist

JoshuaJackson098:
Zrzut ekranu 2026-04-18 185131

Zrzut ekranu 2026-04-18 185211

added:
dwseal.com
sskaid.com


martenson:
@JoshuaJackson098 Thanks for your suggestion but we do not simply merge with other lists. We want to verify the domain is actually used by a disposable provider by ourselves. Do you know where we can obtain a disposable email with this domain? If so please provide a screenshot of the disposable domain being used there.

PR #947, opened 2026-04-18
Added domains seen in 10 minute mail

romh-acking:

https://10minutemail.com/


martenson:
thank you @romh-acking

PR #952, opened 2026-04-22
Add ruutukf.com to disposable email blocklist

VeArnold:
Source:
https://verifymail.io/domain/ruutukf.com

### DIFF
--- a/disposable_email_blocklist.conf
+++ b/disposable_email_blocklist.conf
@@ -3998,6 +3998,7 @@ ruru.be
 rustydoor.com
 rustyload.com
 ruu.kr
+ruutukf.com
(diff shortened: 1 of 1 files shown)


martenson:
@VeArnold Thanks for your suggestion but we do not simply merge with other lists. We want to verify the domain is actually used by a disposable provider by ourselves. Do you know where we can obtain a disposable email with this domain? If so please provide a screenshot of the disposable domain being used there.

VeArnold:
No worries. :) I'll close this then, I'm a good person, but not that good. Thanks for your time.

martenson:
@VeArnold sure, have a good one

PR #955, opened 2026-04-25
Add hidingmail.net

ucarno:
Can be obtained from here: https://adguard.com/en/adguard-temp-mail/overview.html

image


martenson:
thank you @ucarno !

PR #958, opened 2026-04-28
Add tempmail.fish domains

edtaylors:
Added https://tempmail.fish domains
- calmriver.info
- clearbeam.pro
- frostypeak.info


martenson:
thank you @edtaylors !

PR #959, opened 2026-04-30
Add verified temporary, burner, and spam email domains

ADITYA-KUMAR-2358:
Hello Maintainers,

I have found and verified many domains that are being used by disposable emails, burner inboxes, spam-traps, and typo-squatters to avoid registrations.

All these domains have been deliberately chosen and verified through publicly available temp mail database, disposable email generators, spam trap links, and typical typo domains. They give an instant inbox using mirror websites such as Mailinator, GuerrillaMail, 10MinuteMail, YOPMail, MailDrop, 1SecMail, Temp-Mail, Mohmal, etc.

Because of the large quantity, I cannot provide you with screenshots for all the domains, but I am willing to do that from the most prominent providers.

### DIFF
--- a/disposable_email_blocklist.conf
+++ b/disposable_email_blocklist.conf
@@ -2346,6 +2346,7 @@ inoutmail.eu
 inoutmail.info
 inoutmail.net
 inpwa.com
+inreur.com
 insanony.art
 insanony.one
 insanony.store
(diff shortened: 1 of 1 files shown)


martenson:
@ADITYA-KUMAR-2358 wow that's a lot of prose with two domain names 🤣

ADITYA-KUMAR-2358:
@martenson , I only need to add around 400 more websites, but I felt that adding these two was a priority because these are the most commonly used temporary email addresses. I’m also looking for other websites that I can add. Thankyou for your reply.

martenson:
@ADITYA-KUMAR-2358 cool, thanks. Please cut down on the LLM-generated prose in PR body though, it is tiresome to read.

PR #962, opened 2026-05-07
Update disposable email list from temp-mail.org

chirzul:

image


martenson:
thanks @chirzul !

PR #969, opened 2026-05-17
Add wshu.net to disposable email blocklist

BijayaKumarTiadi:
From : https://mail.tm/en/
domain : wshu.net
image


martenson:
thanks @BijayaKumarTiadi !

PR #972, opened 2026-05-18
Add new domainparts for tempmail.id.vn

Crmteam-IVG:
New domainparts for tempmail.id.vn:

- hdiscord.xyz
- nghienplus.store
- uki.io.vn

image


martenson:
thanks a bunch @Crmteam-IVG !

PR #977, opened 2026-05-22
Add ameady.com to blocklist

keshavarun20:
ameady.com is a disposable temp email domain from temp-mail.org.

Verified via check-mail.org 
brave_screenshot_check-mail org
 risk score 91/100, is_disposable: true.
Screenshot attached.

Ran maintain.sh to sort and deduplicate.

### DIFF
--- a/disposable_email_blocklist.conf
+++ b/disposable_email_blocklist.conf
@@ -416,6 +416,7 @@ amazon-aws.org
 amberwe.us
 ambiancewe.us
 ambitiouswe.us
+ameady.com
 amelabs.com
 americanawe.us
 americancivichub.com


jonesdevelopment (approved):

martenson:
@keshavarun20 Thanks for your suggestion but we do not simply merge with other lists. We want to verify the domain is actually used by a disposable provider by ourselves. Do you know where we can obtain a disposable email with this domain? If so please provide a screenshot of the disposable domain being used there.

keshavarun20:
I verified that `ameady.com` is actively used as a disposable email domain by Temp Mail (https://temp.mail). As you can see in the attached screenshot, the domain is being served live on their platform. Please consider adding it to the list.
image

martenson:
thanks @keshavarun20

PR #978, opened 2026-05-23
Add gzeos.com to disposable email blocklist

nomikugg:
Added gzeos.com to the disposable email blocklist.

Evidence:
gzeos.com is used by temp-mail and can generate temporary addresses.

Screenshot attached below.
temp2


martenson:
thanks @nomikugg

PR #992, opened 2026-05-31
Add new domains to disposable email blocklist

surenpoghosian:
Added new disposable email domains to the blocklist.

I am not aware where one generates those emails, but they're spamming my system for a while now.

Screenshot 2026-05-20 at 19 05 03


martenson:
@surenpoghosian This is not a list of "spammer" domains but a list of domains used for identified disposable services. I think qualifications are different for these lists.

PR #993, opened 2026-06-01
added new tinyhost domains

thibaultmol:
image

seems like they added a bunch more domains in the last couple days

### DIFF
--- a/disposable_email_blocklist.conf
+++ b/disposable_email_blocklist.conf
@@ -601,6 +601,7 @@ bdshar.com
 beaconmessenger.com
 bearsarefuzzy.com
 beastmail.space
+beatbok.online
 beddly.com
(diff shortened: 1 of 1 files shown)


thibaultmol:
so the *io.vn domains .... (cause i made this mistake last time as well)
those should have their subdomain added but not io.vn itself?
cause i see other *.io.vn domains in the file

martenson:
@thibaultmol io.vn is likely a [public suffix](https://github.com/disposable-email-domains/disposable-email-domains/blob/f94e2081e81f52f29f8c8138e4c7fb4d087c2313/verify.py#L10) and in that case we add the third level domains that are used by disposable email services

martenson:
thank you @thibaultmol

PR #994, opened 2026-06-02
another set of tinyhost domains

thibaultmol:
They add them faster than I would have imagined
Screenshot_20260602_083016


martenson:
yep, they do...
cheers @thibaultmol

martenson:
this could likely be added to auto fetcher (https://github.com/disposable-email-domains/disposable-email-domains/blob/main/fetch_domains.py)

PR #998, opened 2026-06-03
Added 1t-mail.com

podcroco:
image
This is a Japanese site, but at the URL
https://www.onetime-mail.com/
you can get a disposable email address using ***@1t-mail.com.

Please add it to your list.


martenson:
thank you @podcroco

PR #1003, opened 2026-06-05
Add new disposable email domains to blocklist

abarcelosanchez:
## Domains to include in conf file

**ixospace.com**

**40xkxkud.com**

**devto.biz.id**

Here screenshots evidences about temporary email:

domain-1
domain-2
domain-3

Thanks!

### DIFF
--- a/disposable_email_blocklist.conf
+++ b/disposable_email_blocklist.conf
@@ -144,6 +144,7 @@
 3trtretgfrfe.tk
 4-n.us
 4057.com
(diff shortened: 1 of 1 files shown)


martenson:
@abarcelosanchez Thanks for your suggestion but we do not simply merge with other lists. We want to verify the domain is actually used by a disposable provider by ourselves. Do you know where we can obtain a disposable email with this domain? If so please provide a screenshot of the disposable domain being used there.

PR #1006, opened 2026-06-06

Dhayal-cloud:


martenson:
@Dhayal-cloud Thanks for your suggestion but we want to verify the domain is actually used by a disposable provider by ourselves. Do you know where we can obtain a disposable email with this domain? If so please provide a screenshot of the disposable domain being used there.

PR #1009, opened 2026-06-07
Add atomicmail.io, passmail.net, simplelogin.com

ulyanas:
## Summary

Adds these disposable email domains to `disposable_email_blocklist.conf`:

- `atomicmail.io`
- `passmail.net`
- `simplelogin.com`

## Proofs

`atomicmail.io` offers email alias creation at `atomicmail.io/create-email-alias`:

atomicmail.io proof

`passmail.net` is used by Proton Pass aliases:

passmail.net proof

`simplelogin.com` offers email aliases:

simplelogin.com proof

## Validation

- Ran `./maintain.sh`
- Ran `/tmp/disposable-email-domains-verify-venv/bin/python verify.py`

### DIFF
--- a/disposable_email_blocklist.conf
+++ b/disposable_email_blocklist.conf
@@ -643,6 +643,7 @@ atelierdelune.shop
 atmemail.top
 atnextmail.com
 atombomb.online
+atomicmail.io
 attnetwork.com
 au1688x.us
 aubady.com
@@ -4595,6 +4596,7 @@ parthekio.live
 parthxxtoasts.site
 passeone.com
 passeone.xyz
+passmail.net
 pastebitch.com
 pastryofistanbul.com
 patity.com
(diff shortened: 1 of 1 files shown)


thibaultmol:
Does this count as disposable domains though?
I personally use simplelogin (granted with a custom domain, so this wouldn't affect me) but i'm curious if alias services like this would fall under this repo's purpose or not.

with these you actually need to make an account and they're paid services (atleast proton's ones) so if abuse happens, those can be reported and proton will act accordingly and ban the entire account.

while the usual disposable email services this repo covers are actual "no login needed, just load our page and voila here's a fresh temp email address"

martenson:
@ulyanas thanks for the PR however, alias service is not considered disposable in our view. Look for "mozilla relay" in this repo to discover more of the reasoning.

martenson:
@thibaultmol @nillpoe thank you for the review, it is valued

PR #1012, opened 2026-06-10
Add ulm-dsl.de

NereosRed:
The disposable email adresses can be generated using the form on [ulm-dsl.de](https://ulm-dsl.de/):

grafik


martenson:
@NereosRed thank you

PR #1013, opened 2026-06-10
Remove Dead Domains

SamHimmy:
All the domains are dead have no dns records
co.cc redirects to [https://cc.cc](https://cc.cc) does not give temp mail


martenson:
@SamHimmy I'll need to think about this one. The fact a domain was disposable in the past is likely a pretty good indicator of it being used for the same purpose in the future.

PR #1017, opened 2026-06-12

barbaraborges:
Adding the domains:
- hotkev.com from https://temp-mail.org/en/10minutemail
- gonrr.net from https://10minutemail.com/


martenson:
@barbaraborges Thanks for your suggestion -- please provide a screenshot of the disposable domain being offered by the service.

PR #1019, opened 2026-06-12
Add web-library.net

Mario-SO:
## Summary
- Add `web-library.net` to the disposable email domain blocklist.

## Validation
- Ran `./maintain.sh`.
- Ran `python verify.py` in a temporary virtualenv after installing `requirements.txt`; all domain entries are valid.

## Evidence
Cant find the place where people create these emails from, but many checkers get it as disposable

[Skipsend](https://skipsend.com/check/web-library.net/)

Screenshot 2026-06-12 at 19 26 34

[UserCheck](https://www.usercheck.com/domain/web-library.net)

Screenshot 2026-06-12 at 19 26 56

### DIFF
--- a/disposable_email_blocklist.conf
+++ b/disposable_email_blocklist.conf
@@ -6506,6 +6506,7 @@ wearvault.biz.id
 web-contact.info
 web-ideal.fr
 web-inc.net
+web-library.net
 web-mail.pp.ua
 web2mailco.com
 webclub.infos.st


martenson:
@Mario-SO Thanks for your suggestion but we do not simply merge with other lists. We want to verify the domain is actually used by a disposable provider by ourselves. Do you know where we can obtain a disposable email with this domain? If so please provide a screenshot of the disposable domain being used there.

Mario-SO:
@martenson Hi mate, thanks for answering. Seems like this is the place

https://mail.tm/en/

Screenshot 2026-06-15 at 10 15 22

martenson:
@Mario-SO awesome, thanks

Mario-SO:
Those guys provide an API, so many other vibe-coded sites also offer the same web-library.net domain look, i just discovered it.

This site even has the lovable favicon still XD

https://instanttempmail.org/

Screenshot 2026-06-15 at 10 19 41

PR #1023, opened 2026-06-13
Added domains seen in simplelogin.io

romh-acking:

http://simplelogin.io/
The site is login-gated, so you will have to register with a valid email address to see the email domains.

### DIFF
--- a/disposable_email_blocklist.conf
+++ b/disposable_email_blocklist.conf
@@ -270,9 +270,11 @@
 8127ep.com
 816qs.com
 875888.xyz
+8alias.com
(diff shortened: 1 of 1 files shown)


thibaultmol:
Personally my opinion is the same as with the other issue a couple days ago
https://github.com/disposable-email-domains/disposable-email-domains/pull/1009#issuecomment-4721393799
Proton actually does a decent job of handling abuse and this isn't some website where you create aliases. you actually have to go through the hoops to make an account and such. (you can also not do much with the free tier even)

PR #1027, opened 2026-06-14
Add emailinbo.live

khal3d:
`emailinbo.live` is one of the rotating domains used by [mohmal.com](https://www.mohmal.com/en), a free temporary email service that generates disposable addresses expiring after 45 minutes.


martenson:
thanks @khal3d !

PR #1029, opened 2026-06-15
Added Tinyhost new domain fetcher

thesauravpoddar:
I looked at API docs of Tinyhost so i used https://tinyhost.shop/api/all-domains/  to get all domains when we put this link in browser it gives the list all such disposabel 827 domains 
```
{
  "domains": ["khaibn.com", "trungmetax.com", ...],
  "total": 827,
  "last_updated": "2026-06-15T07:04:10.884599"
} this is the exact output i get when i put the link in browser though. 
```
In 
```
if isinstance(data, dict) and "domains" in data:
    for domain in data["domains"]: 
        if isinstance(domain, str) and domain:
            domains.add(domain.lower().strip()
```
``` data["domains"] ``` will get this output perfectly. It will check id the response is a dict and has `"domains"` key and will loop over 827 domain strings checking if it is a non-empty string and then it will add that to domains set

### DIFF
--- a/fetch_domains.py
+++ b/fetch_domains.py
@@ -229,6 +229,39 @@ class GPTMailFetcher(DomainFetcher):
 
         return domains
 
+    class TinyhostFetcher(DomainFetcher):
+    """Fetcher for `tinyhost shop` disposable email domains"""
+
+    def __init__(self):
+        super().__init__("Tinyhost")
+        self.url = "https://tinyhost.shop/api/all-domains/"
+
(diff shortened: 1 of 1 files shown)


martenson:
thanks a bunch @thesauravpoddar !

martenson:
@thesauravpoddar it's refreshing to see hand-submitted code with indent errors :D

thesauravpoddar:
You're welcome! Glad I could help. Happy to contribute! Let me know if anything else is needed

PR #1032, opened 2026-06-16
Add generator.email fetcher

SamHimmy:
fetches domain from generator.email
generator.email domains list change when you refresh. So it checks the generator.email website multiple times


martenson:
awesome, thanks! @SamHimmy

PR #1068, opened 2026-07-07
Add disposable domains: 10minute.email, 10minmail.org

christian-wr:
Both belong to the "10 Minute Mail" disposable/temporary email service and are currently missing from the blocklist:

- 10minute.email — disposable temp-mail provider (page title "10 Minute Mail
  - Disposable Temporary Email Address for Privacy"); redirects to 10minmail.org.
- 10minmail.org — the redirect target / active service domain of the same provider. The list already had 10minmail.de but not the .org.

Verified live 2026-07-07. Inserted in sorted order (maintain.sh is a no-op).

### DIFF
--- a/disposable_email_blocklist.conf
+++ b/disposable_email_blocklist.conf
@@ -52,8 +52,10 @@
 10mail.xyz
 10minemail.com
 10minmail.de
+10minmail.org
 10minut.com.pl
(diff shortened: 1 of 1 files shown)


martenson:
thanks @christian-wr

PR #1069, opened 2026-07-07
Add new disposable email domains

php4fan:
"To add domains please explain where one can generate a disposable email address which uses them, e.g. in the form of screenshots. Without this information your PR will be closed."

Well do whatever you want, I don't know where they come from, but my forum was flooded with spam from users registering with emails with these domains. You can easily verify that the MX records exist.


martenson:
It is impossible for us to verify these. It is "Do I trust the github account that claims these are disposable?" question without any clues.

PR #1074, opened 2026-07-09
added several domains

thibaultmol:
Screenshot_20260709_184137
Screenshot_20260709_183918
Screenshot_20260709_183702
Screenshot_20260709_183654
Screenshot_20260709_183635

### DIFF
--- a/disposable_email_blocklist.conf
+++ b/disposable_email_blocklist.conf
@@ -187,7 +187,6 @@
 2chmail.net
 2ether.net
 2fdgdfgdfgdf.tk
-2mail.co
 2mail.eu.cc
 2odem.com
 2prong.com
@@ -4224,7 +4223,6 @@ mail1web.org
 mail21.cc
 mail22.club
 mail2rss.org
-mail2tor.co
 mail333.com
 mail4.uk
 mail4qa.com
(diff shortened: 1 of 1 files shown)


martenson:
@thibaultmol I think that onionmail and others that require registration do not fall in our "disposable email service" category. I believe there is a fine, while a bit blurry, line between private and disposable.

thibaultmol:
So you're saying that they fall under the same category as simple login then? (because those are actually meant as relaying services, while these ones are purely meant for anonymous stuff, which honestly I don't really see how that would be used for normal usage and kind of only for malicious usage.

Those services that I've added in this PR are specifically ones that we're noticing high signup rates from various mastodon instances. (The email domains they're using to sign up for with them.) 

While I've never noticed these types of things with the domains of the various aliasing services like Simplelogin and Anondaddy (or whatever those other ones were called.)

martenson:
Anonymous email providers requiring registration are out of scope for our list. I understand your point, but this list is not a silver bullet for all service signup filtering needs.

Does that make sense?

thibaultmol:
I forget the exact scope yeah, removed those

martenson:
@thibaultmol thank you, I appreciate your contributions 🌟

PR #1083, opened 2026-07-14
Add disposable domain: duvips.com

nurmdrafi:
I have found this domain from `https://temp-mail.org/en/` then create pull request for contribution

### DIFF
--- a/disposable_email_blocklist.conf
+++ b/disposable_email_blocklist.conf
@@ -2019,6 +2019,7 @@ duskmail.com
 dusrui.com
 dustinry.com
 duya25446.top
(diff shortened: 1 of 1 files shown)


martenson:
@nurmdrafi Thanks for your suggestion but we want to verify the domain is actually used by a disposable provider by ourselves. Do you know where we can obtain a disposable email with this domain? If so please provide a screenshot of the disposable domain being used there.

nurmdrafi:
Screenshot 2026-07-14 at 5 52 24 PM

martenson:
thanks @nurmdrafi !

PR #1106, opened 2026-07-24
add smailpro.com domains

kavore:

image

smailpro.com


martenson:
thanks @kavore !

kavore:
I will add more domains from smailpro. None of them are listed here :(

PR #1123, opened 2026-08-15
Add missing domains from boomlify.com

Fijxu:
image

Those temp-email addresses have been used to register in some of the services that I host at https://nadeko.net just to abuse them.

Thanks for the list ^^


martenson:
thank you @Fijxu !

PR #1140, opened 2026-08-29
Add CleanTempMail fetcher

### LINKED ISSUE #1090: cleantempmail.com
Hi all -

cleantempmail.com supports a large number of domains for email generation visible on https://cleantempmail.com/domains. It would be great if these could be added to the auto-fetcher somehow but I am unable to contribute the update myself currently.

pttydou:
Fixes #1090.

Adds a CleanTempMail fetcher using the public domains endpoint. It follows the API-provided pagination metadata, normalizes string domains, and rejects failed or malformed responses without retaining partial results. Generated blocklist files are unchanged.

Tests:
- `.\.venv\Scripts\python.exe -m unittest discover -v`
- `$env:PYTHONUTF8='1'; .\.venv\Scripts\python.exe verify.py`


test_fetch_domains.py
```diff
@@ -0,0 +1,134 @@
+import io
+import unittest
+from contextlib import redirect_stderr
+from unittest.mock import patch
+
+from fetch_domains import CleanTempMailFetcher, FETCHERS
+
+
+class FakeResponse:
```
martenson: please drop the test, I don't think this is adding enough value

martenson:
@pttydou thanks for the contribution! Would you mind dropping the test file? Parsing faked responses is not much of value here I fear.

pttydou:
Thanks for the feedback. I've removed the test locally, but I'm sorry for the delay in updating the PR. My computer currently only boots in Safe Mode, and the development and network setup needed to validate the change isn't working reliably there. Repairing it may take some time. I'll update the PR once the machine and test environment are working normally again.

test_fetch_domains.py
pttydou: Done — I removed the test file. I reran the live CleanTempMail-to-PSL smoke check and `python verify.py`; both passed.

martenson:
thank you @pttydou !

PR #1142, opened 2026-08-31
Add current active temp-mail.org domains

kaeldrin-gh:
## Summary

Adds 8 domains currently used by **temp-mail.org** for their disposable/temporary email service:

`dd2car.com`, `fanzher.com`, `mapsguy.com`, `mediseat.com`, `neowd.com`, `prorises.com`, `robustq.com`, `slotbeer.com`

None of these were present in the blocklist (checked against main).

> **Note:** This PR supersedes the previously closed PR #1139 (auto-closed after the contributor accidentally deleted their fork). Same content, same evidence.

## Evidence

- Domains were obtained from temp-mail.org's own mailbox API: `POST https://web2.temp-mail.org/mailbox` returns generated addresses on these domains, e.g. `vagal50271@dd2car.com`, `dover70586@slotbeer.com`, `wicabe2077@fanzher.com`, `capiw15607@mediseat.com`.
- All 8 domains are currently live and accept email — each resolves MX records (`mail.<domain>`).
- Confirmed via repeated sampling over multiple days/IPs (~70+ mailbox generations across 4 days, no additional domains in the pool).
- Verified with the repo's own `python verify.py --fix` (exits clean, 8703 valid entries).

Screenshot (address generated on temp-mail.org using `fanzher.com`):

Screenshot 2026-08-28 200216

PR #1143, opened 2026-08-31
Add TempMailFetcher for automated temp-mail.org pool discovery

kaeldrin-gh:
## Summary

Adds a `TempMailFetcher` to `fetch_domains.py`, so the daily "Fetch new domains" bot can automatically discover temp-mail.org's active (rotating) domain pool.

## Why a new fetcher

temp-mail.org does **not** expose its domain list publicly — the `/domains` endpoint requires an auth secret, and the page loads the pool via obfuscated JS. The only reliable observation method is the legacy mailbox API:

```
POST https://web2.temp-mail.org/mailbox
{"method":"getMailBox","params":{"lang":"en"},"id":1}
→ {"token":"eyJ...","mailbox":"vamesix677@prorises.com"}
```

Each call returns a randomly picked domain from the active pool — the same sampling approach already used by `NoopmailFetcher` (which polls `/api/rd` for random domains).

## Design

- Samples up to 30 mailbox generations, 2s apart
- Stops cleanly on the per-IP rate limit (`TooManyRequestsException`, ~5–10 calls/IP) and returns whatever was collected
- Because the daily run only adds *missing* domains, partial coverage per run is fine — the rest of the pool is picked up by subsequent runs
- Degrades gracefully on blocking (Cloudflare/403): empty set → warning, no crash, other fetchers unaffected

## Verification

- Code compiles and registers correctly in `FETCHERS`
- Standalone identical-logic run collected 6 unique domains in one pass (e.g. `vamesix677@prorises.com`, `yetop17512@robustq.com`, `jajec44474@dd2car.com`, `xikekey145@mapsguy.com`, `womak57769@mediseat.com`, `wojegin146@fanzher.com`)
- See also: PR #1142 (adds the current pool manually; this fetcher makes such submissions automatic in the future)

## Known limitation

Like `NoopmailFetcher` (whose comment notes "Blocks github IP range - run locally via vpn/proxy"), GitHub Actions runners may be geo/IP-blocked by temp-mail.org's Cloudflare setup. If the daily run finds nothing, running locally once per week via VPN/Proxy backfills the pool — after which each new rotation is picked up within a run or two.

### DIFF
--- a/fetch_domains.py
+++ b/fetch_domains.py
@@ -366,6 +366,9 @@ class TempMailFetcher(DomainFetcher):
             for _ in range(30):
                 try:
                     response = post(self.url, json=payload, headers=headers, timeout=15)
+                    # Stop early on hard blocks (Cloudflare 403, auth 401)
+                    if response.status_code in (401, 403):
+                        break
                     # Stop on rate limit: {"errorMessage":"Too Many Request","errorName":"TooManyRequestsException"}
                     if response.status_code == 429 or "TooManyRequests" in response.text:
                         break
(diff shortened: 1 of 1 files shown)


kaeldrin-gh:
Thanks for the review — both points addressed in the latest commit (ac31dca):

1. **Retry hammering**: the 2s delay now always runs between attempts (`sleep` moved to `finally`), so failures can no longer produce 30 rapid retries. Also added an early stop on hard blocks (HTTP 401/403 — Cloudflare/auth) instead of retrying into them.

2. **Misleading warning**: the empty-result message now accurately states the likely causes — rate limiting, IP/Cloudflare blocking, or the endpoint changing — instead of "page structure may have changed."

martenson:
Thank you @kaeldrin-gh !

PR #1148, opened 2026-09-03

andrzejbraun:
Domain used by bad actor to generate random domain and to run Economic Denial of Sustainability type of attack


martenson:
this does not seem like a disposable email provider.

See e.g. discussion in https://github.com/disposable-email-domains/disposable-email-domains/pull/1127

PR #1156, opened 2026-09-07
Add 94an.com to disposable email blocklist

vikasgaur24:
This change adds the domain 94an.com to the disposable email blocklist to prevent temporary addresses from this provider from being accepted.The domain was found active on the temp-mail service, as shown below:

94an com_temp-mail


martenson:
this has already been in the list

PR #1167, opened 2026-09-10
Fix stale Tinyhost and GeneratorEmail fetchers

kaeldrin-gh:
## Summary

Fixes two fetchers that have been returning **0 domains** in the daily run (`Found 0 domains from Tinyhost` / `GeneratorEmail`), restoring two dead sources.

## Tinyhost

- The old `GET /api/all-domains/` endpoint now returns **401 Unauthorized** (requires an `X-Domain-Token`)
- Switched to the public endpoint the site UI uses: `GET /api/random-domains/?page=N&limit=50` → `{"domains":[...],"total":50}`
- Each page returns a random selection from the pool, so the fetcher samples up to 10 pages per run; coverage accumulates across daily runs
- Verified locally: **266 domains** fetched in ~6s (pages keep yielding new domains: +50, +44, +35, +28, +25, +21, +12, +19, +11, +10 over 10 pages)

## GeneratorEmail

- The domain dropdown (`<li>` items + `change_dropdown_list` handlers) no longer exists in the page, so the old parser found nothing
- Each page load now selects a random domain and embeds it in the JS config as `cur_domain:"..."`
- Kept the existing 50 concurrent page loads and now extract `cur_domain` from each
- Verified locally: **41 unique domains** from 50 loads in ~2s
- Sampled domains' MX records point to `emailfake.com` / `email-fake.com`, confirming temp-mail infrastructure
- 9 of 14 domains sampled during investigation were missing from the blocklist — the next bot run should pick them up

## Not fixed (deliberately)

- **Noopmail** — endpoint blocks GitHub Actions IPs (documented in the fetcher itself)
- **CyberTemp** — API returns 403 to GitHub IPs (works from residential connections), so a code fix alone won't help

Tested with Python 3.11: compiles clean, both fetchers registered in `FETCHERS` and returning domains.

### DIFF
--- a/fetch_domains.py
+++ b/fetch_domains.py
@@ -313,7 +313,7 @@ class GeneratorEmailFetcher(DomainFetcher):
         super().__init__("GeneratorEmail")
         self.url = "https://generator.email/"
 
-    def _fetch_once(self, attempt: int) -> Set[str]:
+    def _fetch_once(self, attempt: int, domain_pattern: "re.Pattern") -> Set[str]:
         """Fetch and parse the randomly selected domain from a single page load"""
         domains = set()
         try:
@@ -328,7 +328,7 @@ class GeneratorEmailFetcher(DomainFetcher):
         match = re.search(r'cur_domain:"([^"]+)"', response.text)
         if match:
             domain = match.group(1).lower().strip()
-            if domain:
+            if domain_pattern.match(domain):
                 domains.add(domain)
 
         return domains
@@ -336,9 +336,12 @@ class GeneratorEmailFetcher(DomainFetcher):
     def fetch(self) -> Set[str]:
         """Fetch domains by sampling page loads concurrently (domain rotates per load)"""
         domains = set()
+        domain_pattern = re.compile(
+            r'^([a-zA-Z0-9](?:[a-zA-Z0-9\-]{0,61}[a-zA-Z0-9])?(?:\.[a-zA-Z0-9](?:[a-zA-Z0-9\-]{0,61}[a-zA-Z0-9])?)+)$'
+        )
 
         with ThreadPoolExecutor(max_workers=50) as executor:
-            futures = [executor.submit(self._fetch_once, i) for i in range(50)]
+            futures = [executor.submit(self._fetch_once, i, domain_pattern) for i in range(50)]
             for future in as_completed(futures):
                 domains.update(future.result())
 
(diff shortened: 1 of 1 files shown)


kaeldrin-gh:
Thanks — fixed in 987c59d: extracted `cur_domain` values are now matched against the same domain pattern the previous implementation used, so paths, placeholders, or malformed strings can't enter the domain set.

Re-verified live: the fetcher returns domains and every accepted value passes the hostname pattern.

kaeldrin-gh:
Both points addressed in e592fc2:

1. **Tolerant extraction**: `cur_domain` matching now accepts optional whitespace and either quote style (`cur_domain\s*:\s*["']...["']`). The hostname validation added in 987c59d still applies, so the more tolerant extraction can't pollute the set.
2. **Tinyhost early stop**: paging now stops once a page contributes no new domains (duplicate-only page), avoiding unnecessary requests.

Verified with mocked responses: `cur_domain:"..."`, `cur_domain : '...'`, and `cur_domain:'...'` are all extracted, and a duplicate-only page terminates the Tinyhost loop early. Live runs still return domains from both sources.

kaeldrin-gh:
Addressed in bcb4652:

1. **Early stop**: a single all-duplicate page no longer stops the loop — it now requires 4 consecutive duplicate-only pages before stopping, since random sampling can produce an occasional all-duplicate page while unseen domains remain. The 10-page cap still bounds the number of requests.
2. **Error path**: on a request/JSON error the fetcher now returns immediately with whatever was collected so far, so a transient failure no longer falls through to the misleading "page structure may have changed" warning.

Verified with mocked responses: one duplicate page does not stop sampling (later pages are still fetched), four consecutive duplicates do stop it, and an error after partial collection returns the partial set with only the error message. The GeneratorEmail extraction test still passes, and a live Tinyhost run returned 271 domains in ~5s.

kaeldrin-gh:
Addressed in 85a6fe0:

- Tinyhost now passes `params={"page": page, "limit": 50}` to `get()` instead of building the query string manually (the other locations mentioned don't contain manual query building — the other `params` usage in the file is already compliant).
- A malformed response shape now logs a specific diagnostic and returns whatever was collected, rather than breaking silently.
- The exception message now states how many collected domains are being returned.

Verified with mocks: params are passed correctly (page 1..N), malformed responses stop with a specific message, and the error path is explicit about partial results. Live Tinyhost run: 270 domains in ~4s.

martenson:
@kaeldrin-gh looks good, thank you!

martenson:
seems to work fine on CI: https://github.com/disposable-email-domains/disposable-email-domains/pull/1168

PR #1174, opened 2026-09-15
Add 'cherobe.com' to disposable email blocklist

nishkarsh-github:

cherobe.com was created through https://www.emailondeck.com/  

image


martenson:
Thanks for your suggestion but we want to verify the domain is actually used by a disposable provider by ourselves. Do you know where we can obtain a disposable email with this domain? If so please provide a screenshot of the disposable domain being used there.

